PR #19, opened 2019-06-19
Add tests for key press events - up/down arrow, backspace, escape, delete, insert and F1

bhsubra:
Added tests to verify below key press events:
- UpArrow
- DownArrow
- Backspace
- Escape
- Delete
- Insert 
- F1


build/PackageVersions.targets
```diff
@@ -0,0 +1,10 @@
+<Project>
+  <!--
+    This file updates all of the metadata for PackageReferences used throughout the repo.
+    Because it updates items, it must be imported at the end.  While this file only contains properties,
+    it is a .targets file and imported via Directory.Build.targets.
+  -->
+  <ItemGroup>
+  <PackageReference Update="Moq" Version="4.10.1" PrivateAssets="All" />
```
tlmii: nit: Indentation

src/Microsoft.Repl.Tests/ShellTests.cs
```diff
@@ -0,0 +1,191 @@
...
+            mockConsoleManager.SetupSequence(s => s.ReadKey(cancellationToken))
+                        .Returns(consoleKeyInfo);
+            mockConsoleManager.Setup(s => s.CaretPosition)
+                        .Returns(caretPosition);
+
+            Mock<ICommandHistory> mockCommandHistory = new Mock<ICommandHistory>();
+            mockCommandHistory.Setup(s => s.GetPreviousCommand())
+                        .Returns(previousCommand);
+            mockCommandHistory.Setup(s => s.GetNextCommand())
+            .Returns(nextCommand);
```
tlmii: nit: Indentation (and consistency of indentation) of the .Returns calls.

bhsubra: Fixed here: https://github.com/aspnet/HttpRepl/commit/fb71f0807538c6980ebc937163f89cf70b330f31

Thanks!

src/Microsoft.Repl.Tests/ShellTests.cs
```diff
@@ -0,0 +1,191 @@
...
+            // Verify the input buffer contents before F1 key press event
+            Assert.Equal(inputBufferText, shell.ShellState.InputManager.GetCurrentBuffer());
+
+            shell.RunAsync(cancellationTokenSource.Token);
+
+            // Verify the input buffer contents after F1 key press event
+            Assert.Equal(inputBufferText, shell.ShellState.InputManager.GetCurrentBuffer());
+        }
+
+        private Shell CreateShell(ConsoleKey consoleKey, int caretPosition, string previousCommand, string nextCommand, out CancellationTokenSource cancellationTokenSource)
```
tlmii: What do you think about splitting this function out a bit to make it more clear what's mocked and what's "real"? Right now the httpState, defaultCommandDispatcher, cancellationTokenSource/cancellationToken, shellState and the Shell itself are all "real". The IConsoleManager and ICommandHistory are mocked. I think it might be easier when we come back to this code to add more tests in a few months or whatever if we can at a glance see what's being mocked. I'm not positive its necessary though. Thoughts?

bhsubra: I'll be sending out a pr with tests for the remaining (handled) key press events - LeftArrow, RightArrow, Home, End, Tab and Enter. From my investigation, they are not straightforward and I'm guessing this method needs to be revisited and tweaked to handle those scenarios. I will leave this as is now and will tweak it/refactor as required in my next pr.

Let me know if anyone else has any other suggestions.

src/Microsoft.Repl.Tests/ShellTests.cs
```diff
@@ -0,0 +1,191 @@
...
+        {
+            string previousCommand = "set base \"https://localhost:44366/\"";
+            Shell shell = CreateShell(ConsoleKey.UpArrow,
+                0,
+                previousCommand,
+                null,
+                out CancellationTokenSource cancellationTokenSource);
+
+            // Verify the input buffer is empty before the UpArrow key press event
+            Assert.Equal(string.Empty, shell.ShellState.InputManager.GetCurrentBuffer());
```
jimmylewis: This isn't relevant to the test scenario, is it?

build/PackageVersions.targets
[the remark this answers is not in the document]
bhsubra: Fixed here:
https://github.com/aspnet/HttpRepl/blob/dev/bhsubra/AddUnitTests/build/PackageVersions.targets

src/Microsoft.Repl.Tests/ShellTests.cs
```diff
@@ -0,0 +1,191 @@
...
+namespace Microsoft.Repl.Tests
+{
+    public class ShellTests
+    {
+        [Fact]
+        public void RunAsync_WithUpArrowKeyPress_UpdatesCurrentBufferWithPreviousCommand()
+        {
+            string previousCommand = "set base \"https://localhost:44366/\"";
+            Shell shell = CreateShell(ConsoleKey.UpArrow,
+                0,
```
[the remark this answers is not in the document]
bhsubra: Updated here:
https://github.com/aspnet/HttpRepl/commit/4c256c6c0aa71b9caaa6c64cc91146d885eeb414

src/Microsoft.Repl.Tests/ShellTests.cs
bhsubra: I want to verify buffer contents before and after keypress events.

jimmylewis: After makes sense, but in this test case, it's irrelevant what it is before, right?  

Actually, now that I think about it, for thoroughness you could have a test where there is something on the buffer before and verify that pressing Up/Down replaces it.

src/Microsoft.Repl.Tests/ShellTests.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.Repl.Tests
     public class ShellTests
     {
         [Fact]
-        public void RunAsync_WithUpArrowKeyPress_UpdatesCurrentBufferWithPreviousCommand()
+        public async void RunAsync_WithUpArrowKeyPress_UpdatesCurrentBufferWithPreviousCommand()
```
jimmylewis: async **Task**

src/Microsoft.Repl.Tests/ShellTests.cs
bhsubra: So removed before assert from all tests and added new test to check buffer contents before and after key press event:
https://github.com/aspnet/HttpRepl/commit/2f322167e964e7f08a6b8343ea0943296f22ad84

src/Microsoft.Repl.Tests/Microsoft.Repl.Tests.csproj
```diff
@@ -5,6 +5,11 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Moq" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.HttpRepl\Microsoft.HttpRepl.csproj" />
```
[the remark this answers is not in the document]
bhsubra: Removed the dependency here:
https://github.com/aspnet/HttpRepl/commit/97c3dad53aa681a26f79f7e6d743975d9a0c8afc

src/Microsoft.Repl.Tests/ShellTests.cs
```diff
@@ -162,8 +162,8 @@ public async Task RunAsync_WithUnhandledKeyPress_DoesNothing()
 
         private Shell CreateShell(ConsoleKey consoleKey, int caretPosition, string previousCommand, string nextCommand, out CancellationTokenSource cancellationTokenSource)
         {
-            HttpState httpState = new HttpState();
-            var defaultCommandDispatcher = DefaultCommandDispatcher.Create(httpState.GetPrompt, httpState);
+            MockHttpState mockHttpState = new MockHttpState();
```
jimmylewis: The Http here is a red herring.  In the Microsoft.Repl layer, there's nothing Http specific, and hence no reason to refer to Http.  It's more clear to use something like the [Null Object pattern](https://en.wikipedia.org/wiki/Null_object_pattern) to indicate that this state object has no behavior (which is kinda what you've done but the name is confusing - normally you'd see something like `NullState`).

In this case, we can even go a step further because there is no type required for the state, and also the prompt can be decoupled from the state, so we could just do something like:

```
var defaultCommandDispatcher = DefaultCommandDispatcher.Create(() => string.Empty, new object());
```

bhsubra: Updated here:
https://github.com/aspnet/HttpRepl/commit/c4c9fa26b8157d81e7312facf8490e26642a895b

src/Microsoft.Repl/Input/InputManager.cs
```diff
@@ -201,6 +201,11 @@ public async Task StartAsync(IShellState state, CancellationToken cancellationTo
                 {
                     ConsoleKeyInfo keyPress = state.ConsoleManager.ReadKey(cancellationToken);
 
+                    if (keyPress.Key == 0 && keyPress.KeyChar == '\0')
```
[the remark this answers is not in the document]
bhsubra: Thanks for helping figure this out Jimmy!
Updated here:
https://github.com/aspnet/HttpRepl/commit/e5aea8a75557a718f94d864f7f92479ad6b32cd9

src/Microsoft.Repl.Tests/ShellTests.cs
```diff
@@ -0,0 +1,191 @@
...
+        private Shell CreateShell(ConsoleKey consoleKey, int caretPosition, string previousCommand, string nextCommand, out CancellationTokenSource cancellationTokenSource)
+        {
+            var defaultCommandDispatcher = DefaultCommandDispatcher.Create(x => { }, new object());
+
+            Mock<IConsoleManager> mockConsoleManager = new Mock<IConsoleManager>();
+            var cts = new CancellationTokenSource();
+            CancellationToken cancellationToken = cts.Token;
+            ConsoleKeyInfo consoleKeyInfo = new ConsoleKeyInfo('\0', consoleKey, false, false, false);
+            mockConsoleManager.Setup(s => s.ReadKey(cancellationToken))
+                .Callback(() => cts.Cancel())
```
jimmylewis: nit: might be worth putting a comment here to call out that this is used to break the loop so that shell.RunAsync() returns.

bhsubra: Sure, added here:
https://github.com/aspnet/HttpRepl/commit/8b3c9233ec033b53939f13dfb0cf4cd302f9ff8d

PR #23, opened 2019-06-21
Added tests for remaining handled key press events

bhsubra:
Added tests for below key press events:
- Enter
- Tab 
- Shift + Tab
- Control + U
- Left Arrow
- Ctrl + Left Arrow
- Right Arrow
- Ctrl + Right Arrow
- Home
- Ctrl + A
- End
- Ctrl + E

### DIFF
--- a/src/Microsoft.Repl.Tests/Microsoft.Repl.Tests.csproj
+++ b/src/Microsoft.Repl.Tests/Microsoft.Repl.Tests.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.0</TargetFramework>
--- a/src/Microsoft.Repl.Tests/Mocks/MockConsoleManager.cs
+++ b/src/Microsoft.Repl.Tests/Mocks/MockConsoleManager.cs
@@ -29,9 +29,9 @@ namespace Microsoft.Repl.Tests.Mocks
         {
         }
 
-        public void MoveCaret(int position)
+        public void MoveCaret(int offset)
         {
-            CaretPosition += position;
+            CaretPosition += offset;
         }
 
         public ConsoleKeyInfo ReadKey(CancellationToken cancellationToken)
--- a/src/Microsoft.Repl.Tests/Mocks/MockConsoleManager.cs
+++ b/src/Microsoft.Repl.Tests/Mocks/MockConsoleManager.cs
@@ -1,0 +1,72 @@
+using System;
+using System.Threading;
+using Microsoft.Repl.ConsoleHandling;
+
+namespace Microsoft.Repl.Tests.Mocks
+{
+    internal class MockConsoleManager : IConsoleManager
+    {
+        private CancellationTokenSource _cancellationTokenSource;
+        private ConsoleKeyInfo _consoleKeyInfo;
+
+        public MockConsoleManager(ConsoleKeyInfo consoleKeyInfo, CancellationTokenSource cancellationTokenSource)
+        {
+            _cancellationTokenSource = cancellationTokenSource;
+            _consoleKeyInfo = consoleKeyInfo;
+        }
+
+        public Point Caret => throw new NotImplementedException();
+
+        public Point CommandStart => throw new NotImplementedException();
+
+        public int CaretPosition { get; set; }
+
+        public IWritable Error => new MockWritable();
+
+        public bool IsKeyAvailable => throw new NotImplementedException();
+
+        public void Clear()
+        {
+        }
+
+        public void MoveCaret(int position)
+        {
+            CaretPosition += position;
+        }
+
+        public ConsoleKeyInfo ReadKey(CancellationToken cancellationToken)
+        {
+            _cancellationTokenSource.Cancel();
+            return _consoleKeyInfo;
+        }
+
+        public void ResetCommandStart()
+        {
+        }
+
+        public IDisposable AddBreakHandler(Action onBreak)
+        {
+            return null;
+        }
+
+        public void Write(char c)
+        {
+        }
+
+        public void Write(string s)
+        {
+        }
+
+        public void WriteLine()
+        {
+        }
+
+        public void WriteLine(string s)
+        {
+        }
+
+        public bool IsCaretVisible { get => true; }
+
+        bool IWritable.IsCaretVisible { get => true; set => value = true; }
+    }
+}
--- a/src/Microsoft.Repl.Tests/ShellTests.cs
+++ b/src/Microsoft.Repl.Tests/ShellTests.cs
@@ -240,15 +240,17 @@ namespace Microsoft.Repl.Tests
                 nextCommand: null,
                 out CancellationTokenSource cancellationTokenSource);
 
-            string inputBufferTextBeforeKeyPress = "g";
-            string inputBufferTextAfterKeyPress = "get";
+            string inputBufferTextBeforeKeyPress = "c";
+            string inputBufferTextAfterKeyPress = "cd";
 
             IShellState shellState = shell.ShellState;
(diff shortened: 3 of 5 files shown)


src/Microsoft.Repl.Tests/Microsoft.Repl.Tests.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
```
bhsubra: Not sure why this is showing up. I didn't touch this file

src/Microsoft.Repl.Tests/ShellTests.cs
```diff
@@ -187,5 +494,12 @@ private Shell CreateShell(ConsoleKey consoleKey, int caretPosition, string previ
 
             return new Shell(shellState);
         }
+
+        private void VerifyMoveCaretMethodWasCalledOnce(IShellState shellState, int caretPosition, Times times)
```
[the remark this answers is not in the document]
bhsubra: Yes, I forgot that updated this method. Thanks!

Updated here:
https://github.com/aspnet/HttpRepl/commit/8c8409a4f9ff4b4c71e2219e45ef5a65e322bcc7

src/Microsoft.Repl.Tests/ShellTests.cs
```diff
@@ -14,7 +15,8 @@ public class ShellTests
         public async Task RunAsync_WithUpArrowKeyPress_UpdatesCurrentBufferWithPreviousCommand()
         {
             string previousCommand = "set base \"https://localhost:44366/\"";
-            Shell shell = CreateShell(consoleKey: ConsoleKey.UpArrow,
+            ConsoleKeyInfo consoleKeyInfo = new ConsoleKeyInfo('\0', ConsoleKey.UpArrow, false, false, false);
```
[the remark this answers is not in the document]
bhsubra: Yup, that would make it more readable. Updated here:
https://github.com/aspnet/HttpRepl/commit/dc16f46340247ba1f94a36a58fdee157c3f9921a

src/Microsoft.Repl.Tests/ShellTests.cs
```diff
@@ -159,19 +190,295 @@ public async Task RunAsync_WithUnhandledKeyPress_DoesNothing()
...
+            shellState.InputManager.SetInput(shellState, inputBufferContents);
+
+            await shell.RunAsync(cancellationTokenSource.Token).ConfigureAwait(false);
+
+            // The first method call is from SetInput(..) and the second one is from End key press event
+            VerifyMoveCaretMethodWasCalledOnce(shellState, -3, Times.Exactly(2));
+        }
+
+        [Fact]
+        public async Task RunAsync_WithCtrlAKeyPress_VerifyMoveCaretWasCalled()
```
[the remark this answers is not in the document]
bhsubra: It's supposed to have the same behavior as Home key https://github.com/aspnet/HttpRepl/blob/master/src/Microsoft.Repl/Input/KeyHandlers.cs#L21

Tim helped figure this out. It behaves the same as home key if control key shortcuts in the console options is disabled. If the option is on,  Ctrl + A is caught by the console itself and it highlights the whole console window. It doesn't get to the http repl processor

I think it works as expected in test as the ConsoleManager is mocked.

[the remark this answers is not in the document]
tlmii: Yep, it seems to work as expected with that setting turned off, and in other consoles (like Ubuntu on WSL's bash console). So we're good there.

src/Microsoft.Repl.Tests/ShellTests.cs
```diff
@@ -159,19 +226,347 @@ public async Task RunAsync_WithUnhandledKeyPress_DoesNothing()
...
+            defaultCommandDispatcher.AddCommand(new MockCommand(commandName));
+
+            await shell.RunAsync(cancellationTokenSource.Token).ConfigureAwait(false);
+
+            // Verify the input buffer contents after tab key press event
+            Assert.Equal(inputBufferTextAfterKeyPress, shell.ShellState.InputManager.GetCurrentBuffer());
+        }
+
+        [Fact]
+        public async Task RunAsync_WithShiftTabKeyPress_UpdatesInputBufferWithFirstEntryFromSuggestionList()
```
jimmylewis: Tab and Shift+Tab do the same thing?

src/Microsoft.Repl.Tests/ShellTests.cs
```diff
@@ -159,19 +226,347 @@ public async Task RunAsync_WithUnhandledKeyPress_DoesNothing()
...
+                out CancellationTokenSource cancellationTokenSource);
+
+            string inputBufferContents = "set base \"https://localhost:44366/\"";
+
+            IShellState shellState = shell.ShellState;
+            shellState.InputManager.SetInput(shellState, inputBufferContents);
+
+            await shell.RunAsync(cancellationTokenSource.Token).ConfigureAwait(false);
+
+            VerifyMoveCaretMethodWasCalled(shellState, -1, Times.Once());
```
jimmylewis: I dislike this validation approach.
1. It tests an implementation detail (was a method called a specified number of times with a given parameter).  This is inflexible to changes in the implementation that might set the caret position differently.  What if it were called with a different parameter? A different number of times? What if the implementation set the caret position directly instead of moving it by an offset?  Most importantly, should this test fail if any of these change?
2. It doesn't validate that the caret is in the expected position after the action is taken.

IMO the test should be asserting the caret position (the state of the system), not how it was moved.

bhsubra: I am using a mock of IConsoleManager. So can't verify the caret position. This is the best I could come up with. Can we please discuss this offline when you have time? Thanks!

src/Microsoft.Repl.Tests/ShellTests.cs
bhsubra: Actually it should be first and last. Since I had only one command, it behaved the same. I updated the inputs here to be clear:
https://github.com/aspnet/HttpRepl/commit/c7e2dcc3a354fb3ad79797208be7415e8548b725

src/Microsoft.Repl.Tests/ShellTests.cs
bhsubra: @jimmylewis , as discussed offline, made changes here:
https://github.com/aspnet/HttpRepl/commit/f3eeb8fa792d5435e3712f073b06f1c0a3190f51

Could you please take a look and let me know what you think. Thanks!

jimmylewis: I like the approach, but there's some issues with it. Can you get that commit into this PR? (If I comment on the commit, they won't show up in any PR).

For example:
        public void MoveCaret(int position)
        {
            CaretPosition = position;
        }
`position` should be `offset`, and it should modify position but not overwrite it.  In the tests, you're asserting that the caret has a negative position - that doesn't make sense.  For example, when you start at position 3 and then go to the beginning, the test should assert that the caret position is now 0, not -3.

bhsubra: Oops, sorry. Looks like I committed to wrong branch. Updated commit:
https://github.com/aspnet/HttpRepl/commit/cec25199e0d8fb991c1856832205ff381318af36

src/Microsoft.Repl.Tests/Mocks/MockConsoleManager.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        public IWritable Error => new MockWritable();
+
+        public bool IsKeyAvailable => throw new NotImplementedException();
+
+        public void Clear()
+        {
+        }
+
+        public void MoveCaret(int position)
```
jimmylewis: Can you rename this to `offset` instead of `position`, including in the IConsoleManager interface.  Using `position` is a confusing name because it sounds like "MoveCaret To Position", not "MoveCaret By Offset" (which is what both implementations actually do).

PR #24, opened 2019-06-21
First pieces of GET/POST tests

tlmii:


src/Microsoft.HttpRepl.Tests/Commands/GetCommandTests.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            GetCommand command = new GetCommand();
+            ICoreParseResult parseResult = CoreParseResultHelper.Create("GET");
+            CancellationTokenSource cts = new CancellationTokenSource();
+
+            await command.ExecuteAsync(shellState, httpState, parseResult, cts.Token);
+
+            Assert.Equal(expectedResponseLines, actual.Count); 
+            Assert.Equal(expectedResponseContent, actual[expectedResponseLines - 1]);
+
```
bhsubra: nit: Empty line after Assert here and next method

src/Microsoft.HttpRepl.Tests/Commands/GetCommandTests.cs
```diff
@@ -0,0 +1,119 @@
...
+            ICoreParseResult parseResult = CoreParseResultHelper.Create("GET");
+            CancellationTokenSource cts = new CancellationTokenSource();
+
+            await command.ExecuteAsync(shellState, httpState, parseResult, cts.Token);
+
+            Assert.Equal(expectedErrorMessage, actualErrorMessage);
+        }
+
+        [Fact]
+        public async Task MultipartRoute()
```
bhsubra: Per Jimmy's suggestion in one of the code reviews, I've been trying to follow this naming scheme - <MethodBeingTested>_<DescribeInputScenario>_<DescribeExpectedResult>. It makes the test name clear and it's easier to tell what scenario is being tested without looking at the test implementation.

src/Microsoft.HttpRepl.Tests/Microsoft.HttpRepl.Tests.csproj
```diff
@@ -7,6 +7,11 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Moq" Version="4.10.1" />
```
bhsubra: Do we need the version here? Shouldn't all the versioning information be in build\PackageVersions.targets?

src/Microsoft.HttpRepl/Program.cs
```diff
@@ -9,6 +9,9 @@
 using Microsoft.Repl.ConsoleHandling;
 using Microsoft.Repl.Parsing;
 using Microsoft.HttpRepl.Commands;
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Microsoft.HttpRepl.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
```
bhsubra: Can we create a Properties\AssemblyInfo.cs file and move this there?

src/Microsoft.HttpRepl/StringResources.cs
```diff
@@ -0,0 +1,7 @@
+namespace Microsoft.HttpRepl
+{
+    internal static class StringResources
+    {
+        public static readonly string Error_NoBasePath = "'set base {url}' must be called before issuing requests to a relative path";
```
bhsubra: This needs to be localized in future. Should we put it in a Strings.Resx file instead?

src/Microsoft.HttpRepl.Tests/Commands/GetCommandTests.cs
tlmii: Removed in e3302fb

src/Microsoft.HttpRepl.Tests/Commands/GetCommandTests.cs
tlmii: Changed in e3302fb

src/Microsoft.HttpRepl.Tests/Microsoft.HttpRepl.Tests.csproj
tlmii: I swear I have removed that three or four times. I must be doing something that is putting it back... Removed again in e3302fb

src/Microsoft.HttpRepl/Program.cs
tlmii: Done in e3302fb

src/Microsoft.HttpRepl/StringResources.cs
tlmii: Done in e3302fb.

src/Microsoft.HttpRepl.Tests/Commands/DeleteCommandsTest.cs
```diff
@@ -0,0 +1,110 @@
...
+        private readonly DeleteCommandsConfig _config;
+        public DeleteCommandTests(DeleteCommandsFixture DeleteCommandsFixture)
+        {
+            _config = DeleteCommandsFixture.Config;
+        }
+
+        [Fact]
+        public async Task ExecuteAsync_WithNoBasePath_VerifyError()
+        {
+            HttpState httpState = new HttpState();
```
bhsubra: Hey Tim, the tests here and the ones added before, seem to either verify error/outputs. Except the commandText and command itself, they all pretty much look the same. Do you think we can create a base class and put the Verify(..) methods there? That way we won't be duplicating the code for the other tests.

src/Microsoft.HttpRepl.Tests/Commands/DeleteCommandsTest.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            string expectedErrorMessage = Resources.Strings.Error_NoBasePath.SetColor(httpState.ErrorColor);
+            string actualErrorMessage = null;
+
+            IShellState shellState = MockHelpers.GetMockedShellState(errorMessageCallback: (s) => actualErrorMessage = s);
+
+            DeleteCommand command = new DeleteCommand();
+            ICoreParseResult parseResult = CoreParseResultHelper.Create("DELETE");
+            CancellationTokenSource cts = new CancellationTokenSource();
```
jimmylewis: nit: since you're not cancelling this token anywhere, could could use `CancellationToken.None` instead of declaring the CTS.

src/Microsoft.HttpRepl.Tests/Commands/DeleteCommandsTest.cs
tlmii: Yeah, I initially thought they would be more different but now that I've added some more, it needs a refractor. I'll give it a shot.

src/Microsoft.HttpRepl.Tests/Commands/DeleteCommandsTest.cs
tlmii: Done.

src/Microsoft.HttpRepl.Tests/Commands/DeleteCommandsTest.cs
```diff
@@ -0,0 +1,107 @@
...
+            ICoreParseResult parseResult = CoreParseResultHelper.Create("DELETE");
+
+            await command.ExecuteAsync(shellState, httpState, parseResult, CancellationToken.None);
+
+            Assert.Equal(expectedResponseLines, actual.Count);
+            Assert.Equal(expectedResponseContent, actual[expectedResponseLines - 1]);
+        }
+    }
+
+    public class DeleteCommandsFixture : IDisposable
```
jimmylewis: These fixtures are all the same but for the Config type (as far as I can tell).  Could we have something like this to consolidate them?
```
class TestCommandFixture<TConfigType> 
    where TConfigType : SampleApiServerConfig
```
Then the tests would be something like:
```
public class DeleteCommandTests : IClassFixture<TestCommandFixture<DeleteCommandsConfig>>
```

src/Microsoft.HttpRepl.Tests/Commands/MockHelpers.cs
```diff
@@ -0,0 +1,39 @@
...
+using Microsoft.Repl.ConsoleHandling;
+using Microsoft.Repl.Input;
+using Microsoft.Repl.Suggestions;
+using Moq;
+
+namespace Microsoft.HttpRepl.Tests.Commands
+{
+    internal class MockHelpers
+    {
+        internal static IShellState GetMockedShellState(Action<string> writeLineCallback = null, Action<string> errorMessageCallback = null)
```
jimmylewis: This seems pretty close to already providing a fake implementation of a MockShellState.  You could always have the WriteLineCallback and ErrorMessageCallback exposed as properties, so that tests that use them don't have to set them up with the closure, etc...  just utilize them when needed.

```
var testState = new MockTestState();
...
Assert.Equal(expectedResponseContent, testState.ResponseContent[expectedResponseLines - 2]);
```

src/Microsoft.HttpRepl.Tests/Commands/DeleteCommandsTest.cs
tlmii: This has been pretty heavily refactored now. Take a look and see what you think.

src/Microsoft.HttpRepl.Tests/Commands/DeleteCommandsTest.cs
tlmii: Good call. Done.

src/Microsoft.HttpRepl.Tests/Commands/DeleteCommandsTests.cs
```diff
@@ -0,0 +1,65 @@
...
+                                 baseAddress: _config.BaseAddress,
+                                 path: "a/file/path.txt",
+                                 expectedResponseLines: 5,
+                                 expectedResponseContent: "File path delete received successfully.");
+        }
+
+        [Fact]
+        public async Task ExecuteAsync_WithOnlyBaseAddress_VerifyOutput()
+        {
+
```
bhsubra: nit: empty line. Also can we keep the formatting consistent with above?

src/Microsoft.HttpRepl.Tests/Commands/BaseHttpCommandTests.cs
```diff
@@ -0,0 +1,85 @@
...
+
+            IShellState shellState = MockHelpers.GetMockedShellState(writeLineCallback: (s) => actualResponseContent.Add(s));
+
+            ICoreParseResult parseResult = CoreParseResultHelper.Create(commandText);
+
+            await command.ExecuteAsync(shellState, httpState, parseResult, CancellationToken.None);
+
+            Assert.Equal(expectedResponseLines, actualResponseContent.Count);
+            Assert.Equal(expectedResponseContent, actualResponseContent[expectedResponseLines - 1]);
+
```
bhsubra: nit: empty line

src/Microsoft.HttpRepl.Tests/Commands/DeleteCommandsTest.cs
bhsubra: It looks good to me. Thank you!

src/Microsoft.HttpRepl.Tests/Commands/BaseHttpCommandTests.cs
tlmii: Done in a29a3df

src/Microsoft.HttpRepl.Tests/Commands/DeleteCommandsTests.cs
tlmii: Done in a29a3df

src/Microsoft.HttpRepl.Tests/Commands/MockHelpers.cs
tlmii: This is done now.

src/Microsoft.HttpRepl.IntegrationTests/Microsoft.HttpRepl.IntegrationTests.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Moq" />
+    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
```
jimmylewis: Can you move this package version into the .targets file that @bhsubra added, so that all the package versions are kept in sync?  Should also be cleaned up in the other .csproj as well.

tlmii: Done in 0cf3c64

PR #30, opened 2019-06-25
Add tests for EndpointMetadataReader

bhsubra:
- Added unit tests for EndpointMetadataReader
- Updated SwaggerV1EndpointMetadataReader ReadMetadata(..) to handle array for apis


src/Microsoft.HttpRepl/OpenApi/SwaggerV1EndpointMetadataReader.cs
```diff
@@ -24,9 +24,9 @@ public IEnumerable<EndpointMetadata> ReadMetadata(JObject document)
                 globalConsumes = new JArray();
             }
 
-            if (document["apis"] is JObject obj)
+            if (document["apis"] is JArray jArray)
             {
-                foreach (JProperty property in obj.Properties())
+                foreach (JObject obj in jArray)
```
bhsubra: I mostly modified this based on example I found here:  https://github.com/swagger-api/swagger-codegen/blob/4607a90d7b69463a0ae8fc94fac68fc95d80965e/modules/swagger-codegen/src/test/resources/1_2/petstore-1.2/pet. I don't have a lot of context on what the schema should look like. Let me know your thoughts

jimmylewis: Seems like the best way would be to create a site with a Swagger v1 endpoint and try it out, no?  You can validate the expected schema and get some manual validation of your changes at the same time.

src/Microsoft.HttpRepl.Tests/OpenApi/EndpointMetadataReaderTests.cs
```diff
@@ -0,0 +1,495 @@
+using System.Collections.Generic;
```
jimmylewis: nit: copy the copyright header from other source files

src/Microsoft.HttpRepl.Tests/OpenApi/EndpointMetadataReaderTests.cs
```diff
@@ -0,0 +1,495 @@
...
+using Xunit;
+
+namespace Microsoft.HttpRepl.Tests.OpenApi
+{
+    public class EndpointMetadataReaderTests
+    {
+        [Fact]
+        public void Read_WithInvalidJson_ReturnsNull()
+        {
+            string json = @"{
```
jimmylewis: nit: can you add a comment explaining what is invalid about this JSON payload?  I'm *assuming* it's invalid because it doesn't have a `"swagger"` property.

src/Microsoft.HttpRepl.Tests/OpenApi/EndpointMetadataReaderTests.cs
```diff
@@ -0,0 +1,495 @@
...
+            ""description"": ""Success""
+          }
+        }
+      }
+    }
+  }
+}";
+            List<EndpointMetadata> endpointMetadata = GetEndpointMetadataList(json);
+            IReadOnlyDictionary<string, IReadOnlyDictionary<string, IReadOnlyList<Parameter>>> availableRequests = endpointMetadata[0].AvailableRequests;
+            KeyValuePair<string, IReadOnlyDictionary<string, IReadOnlyList<Parameter>>> firstRequest = availableRequests.First();
```
jimmylewis: Is the enumeration order of the dictionary guaranteed to be consistent?  I.e. will you always get the same results for `.First()` and `.ElementAt(1)`?

src/Microsoft.HttpRepl.Tests/OpenApi/EndpointMetadataReaderTests.cs
```diff
@@ -0,0 +1,495 @@
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.HttpRepl.OpenApi;
+using Newtonsoft.Json.Linq;
+using Xunit;
+
+namespace Microsoft.HttpRepl.Tests.OpenApi
+{
+    public class EndpointMetadataReaderTests
```
jimmylewis: This feels like a suite of tests covering `OpenApiV3EndpointMetadataReader`, `SwaggerV2EndpointMetadataReader`, and `SwaggerV1EndpointMetadataReader`, but there's no validation of the intricacies of the responses (e.g. parameter details).  Should these be split out into tests for the respective metadata readers?

jimmylewis: Similarly, I don't see testing here around registering an additional reader.

src/Microsoft.HttpRepl.Tests/OpenApi/EndpointMetadataReaderTests.cs
tlmii: In practice, probably, but the [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2) says it is non-deterministic.

> For purposes of enumeration, each item in the dictionary is treated as a KeyValuePair<TKey,TValue> structure representing a value and its key. The order in which the items are returned is undefined.

src/Microsoft.HttpRepl.Tests/OpenApi/EndpointMetadataReaderTests.cs
bhsubra: Added here:
https://github.com/aspnet/HttpRepl/commit/6f329922585d50ba40849b897f9da2bbf50fa4f7

src/Microsoft.HttpRepl.Tests/OpenApi/EndpointMetadataReaderTests.cs
bhsubra: Split the tests here:
https://github.com/aspnet/HttpRepl/commit/e95a6c7efc23b60ce7efccda7c49fbf286d3e68c

src/Microsoft.HttpRepl.Tests/OpenApi/OpenApiV3EndpointMetadataReaderTests.cs
```diff
@@ -0,0 +1,215 @@
...
+        public void CanHandle_WithNoOpenApiKeyInDocument_ReturnsFalse()
+        {
+            string json = @"{
+  ""info"": {
+    ""version"": ""v1""
+  },
+   ""paths"": {
+  }
+}";
+            Assert.False(CanHandle(json));
```
jimmylewis: Please keep tests in AAA form for better readability.

jimmylewis:
Will you add a test for EndpointMetadataReader.RegisterReader?  E.g. register a custom fake reader, .Read(), then see if the fake was used.

src/Microsoft.HttpRepl.Tests/OpenApi/EndpointMetadataReaderTests.cs
bhsubra: Updated the test name to Read_WithJObjectFormatNotSupportedByAnyExistingReader_ReturnsNull here: https://github.com/aspnet/HttpRepl/commit/e95a6c7efc23b60ce7efccda7c49fbf286d3e68c

Let me know if it's not clear enough.

Thanks!

src/Microsoft.HttpRepl.Tests/OpenApi/EndpointMetadataReaderTests.cs
bhsubra: Updated to verify dictionary contains key instead:
https://github.com/aspnet/HttpRepl/commit/35818741ff04b0665aacfffe40103dbc9635a00b

src/Microsoft.HttpRepl.Tests/OpenApi/OpenApiV3EndpointMetadataReaderTests.cs
bhsubra: Updated here:
https://github.com/aspnet/HttpRepl/commit/9ca88d9302e5c4830970960b32d34f99654c656b

src/Microsoft.HttpRepl.Tests/OpenApi/EndpointMetadataReaderTests.cs
```diff
@@ -23,5 +26,28 @@ public void Read_WithJObjectFormatNotSupportedByAnyExistingReader_ReturnsNull()
 
             Assert.Null(endpointMetadataReader.Read(jobject));
         }
+
+        [Fact]
+        public void RegisterReader_AddNewReader_VerifyReadReturnsEndpointMetadataCollection()
+        {
+            string json = @"{
+  ""version"": ""1.0.0"",
```
jimmylewis: Instead of using "version" (which is kind of agnostic), what about using "fakeApi" so that it's more clear that this would only apply to the fake reader?

bhsubra: fakeApi sounds better. Thanks!

Updated here:
https://github.com/aspnet/HttpRepl/commit/656236743ed10c9b14dd1b04eecfec61cb1b53af

src/Microsoft.HttpRepl/OpenApi/SwaggerV1EndpointMetadataReader.cs
bhsubra: I spent sometime on this today. I couldn't find a good documentation on how to do this. I'll look into it again over the weekend.

jimmylewis: If you're not confident on it, can we file an issue for follow-up/further investigation?  And then I'm fine with either taking this part of the change (because that's what the example indicates it should be) or reverting it (because keeping the original is a safe play).

bhsubra: Submitted new issue here: https://github.com/aspnet/HttpRepl/issues/92 for further investigation. Thanks!

PR #31, opened 2019-06-26
Abstracting away the file system so it is more testable, and adding tests for it

tlmii:
This includes adding new tests for the PUT and PATCH commands.


src/Microsoft.HttpRepl.IntegrationTests/Commands/PutCommandTests.cs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        public PutCommandsConfig()
+        {
+            Port = SampleApiServerPorts.PutCommandTests;
+            Routes.Add(new DynamicSampleApiServerRoute("PUT", "", RespondWithBody));
+            Routes.Add(new DynamicSampleApiServerRoute("PUT", "this/is/a/test/route", RespondWithBody));
+            Routes.Add(new DynamicSampleApiServerRoute("PUT", "no/body/required", RespondWithBody));
+        }
+
+        private async Task RespondWithBody(HttpContext context)
```
bhsubra: This looks pretty much same in every test. Is there a way to not duplicate it?

src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
```diff
@@ -224,10 +233,10 @@ protected override async Task ExecuteAsync(IShellState shellState, HttpState pro
...
-            await HandleResponseAsync(programState, commandInput, shellState.ConsoleManager, response, programState.EchoRequest, headersTarget, bodyTarget, cancellationToken).ConfigureAwait(false);
+            await HandleResponseAsync(programState, commandInput, shellState.ConsoleManager, response, programState.EchoRequest, headersTarget, bodyTarget, _fileSystem, cancellationToken).ConfigureAwait(false);
         }
 
-        private static async Task HandleResponseAsync(HttpState programState, DefaultCommandInput<ICoreParseResult> commandInput, IConsoleManager consoleManager, HttpResponseMessage response, bool echoRequest, string headersTargetFile, string bodyTargetFile, CancellationToken cancellationToken)
+        private static async Task HandleResponseAsync(HttpState programState, DefaultCommandInput<ICoreParseResult> commandInput, IConsoleManager consoleManager, HttpResponseMessage response, bool echoRequest, string headersTargetFile, string bodyTargetFile, IFileSystem fileSystem, CancellationToken cancellationToken)
```
bhsubra: Does this need to be passed in? Can we not use _fileSystem directly?

src/Microsoft.HttpRepl/Program.cs
```diff
@@ -2,21 +2,49 @@
...
     class Program
     {
         static async Task Main(string[] args)
         {
-            var state = new HttpState();
+            var serviceProvider = new ServiceCollection()
+                .AddSingleton<IFileSystem, RealFileSystem>()
+                .AddSingleton<ICommand<HttpState, ICoreParseResult>, ChangeDirectoryCommand>()
+                .AddSingleton<ICommand<HttpState, ICoreParseResult>, ClearCommand>()
+                //.AddSingleton<ICommand<HttpState, ICoreParseResult>, ConfigCommand>()
```
bhsubra: Is this intentional?

bhsubra: Nvm, ignore this. Looks like it was already commented out.

src/Microsoft.HttpRepl/FileSystem/IFileSystem.cs
```diff
@@ -0,0 +1,17 @@
+using System.Collections.Generic;
+using System.IO;
+
+namespace Microsoft.HttpRepl.FileSystem
+{
+    public interface IFileSystem
+    {
+        Stream Create(string path);
```
jimmylewis: Names of the methods should indicate whether they refer to file or directory operations.

```
CreateFile
DeleteFile
FileExists
etc...
```

I'm surprised you don't have any directory methods in here.  You can't create a file if its parent doesn't exist, is that supposed to be handled implicitly in the IFileSystem implementation, or should it be exposed explicitly from the IFileSystem?
Example:
```
fileSystem.Create(@"C:\doesnotexist\file.txt");
```
Does this throw (because File.Create would), or should the implementation implicitly call `Directory.CreateDirectory(Path.GetDirectoryName(path))`?  Or should IFileSystem expose a CreateDirectory method that must be called first (just like using the System.IO APIs)?

src/Microsoft.HttpRepl.IntegrationTests/Mocks/MockedFileSystem.cs
```diff
@@ -0,0 +1,127 @@
...
+using Microsoft.HttpRepl.FileSystem;
+using Microsoft.HttpRepl.IntegrationTests.Commands;
+
+namespace Microsoft.HttpRepl.IntegrationTests.Mocks
+{
+    internal class MockedFileSystem : IFileSystem
+    {
+        private readonly Dictionary<string, MockedFile> _files = new Dictionary<string, MockedFile>()
+        {
+            // Setup the "files" that should exist in the file system for the various tests here.
```
jimmylewis: Tests should set up the files they expect.  The mock should expose an API for priming the file system for tests.

jimmylewis: Maybe something fluent like:
```
var fileSystem = new MockFileSystem()
                  .AddFile(@"file/path", "content")
                  .AddFile(@"another/path", new byte[] { /* data */ })

File: src/Microsoft.HttpRepl/FileSystem/IFileSystem.cs (hunk 4 above)
Comment (tlmii) [reply src/Microsoft.HttpRepl/FileSystem/IFileSystem.cs]: For the names, those started off in different classes (one for File, one for Path and one for Directory) but there were so few I just moved them together and forgot to rename them. I'll change that.
For the directory stuff... so far I was just focused on adding what was already used, and the existing program doesn't check the directory. Should I harden those calls in this PR as well?

File: src/Microsoft.HttpRepl.IntegrationTests/Mocks/MockedFileSystem.cs [hunk 6]
```diff
@@ -0,0 +1,127 @@
...
+            byte[] bytes = new byte[9];
+            rng.GetBytes(bytes);
+            string path = Convert.ToBase64String(bytes);
+            return path;
+        }
+    }
+
+    internal class MockedFile
+    {
+        public MemoryStream Stream { get; private set; }
```
jimmylewis: You're gonna have some isuses with a Stream as your backing store.  For example, this fails:

    var fs = new MockedFileSystem();
    using (Stream s = fs.Create(@"C:\doesnotexist\file.txt"))
    using (StreamWriter sw = new StreamWriter(s))
    {
        sw.WriteLine("test");
    }

because the stream is not expandable.  You'd also run into issues if the stream was closed and then tried to be read from again later.

jimmylewis: One approach: don't expose Stream from the IFileSystem - do everything with content (`string` or `byte[]`).

If you need to expose Stream, then the mock should probably use a `byte[]` buffer as a backing store, and open a stream to that - it'll give you a block of space to write to for tests, just don't go over the buffer size.

src/Microsoft.HttpRepl/FileSystem/IFileSystem.cs
jimmylewis: It would make sense to harden them (whether changing RealFileSystem to check first, or adding Directory.Exists/Directory.Create wrappers).  From the perspective of IFileSystem as a standalone component, this is a dangerous cliff to fall off - better to have a more robust component now.

src/Microsoft.HttpRepl.IntegrationTests/Mocks/MockedFileSystem.cs
tlmii: If I can get rid of the usage of Stream, that'd definitely make things simpler - that one function (Create) added a lot of complexity to that implementation. I'll take a look and see what that'd look like.

src/Microsoft.HttpRepl.IntegrationTests/Commands/PutCommandTests.cs
tlmii: It's the same for now between PUT/POST/PATCH, but I assume they'll diverge as we add more tests for the specific types of calls. If not, I'll come back and clean that up.

src/Microsoft.HttpRepl.IntegrationTests/Commands/DeleteCommandsTests.cs
```diff
@@ -1,14 +1,19 @@
...
 namespace Microsoft.HttpRepl.IntegrationTests.Commands
 {
     public class DeleteCommandTests : HttpCommandTests<DeleteCommand>, IClassFixture<HttpCommandsFixture<DeleteCommandsConfig>>
     {
+        private static readonly IFileSystem _fileSystem = new MockedFileSystem();
+
         private readonly DeleteCommandsConfig _config;
-        public DeleteCommandTests(HttpCommandsFixture<DeleteCommandsConfig> deleteCommandsFixture)
+        
+        public DeleteCommandTests(HttpCommandsFixture<DeleteCommandsConfig> deleteCommandsFixture) : base(new DeleteCommand(_fileSystem))
```
jimmylewis: nit: put `: base(...)` on the next line for readability

src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
tlmii: It needed to be passed because the method was static. But it didn't need to be static, so I changed it to be an instance method and took out the parameter.

src/Microsoft.HttpRepl.IntegrationTests/Commands/DeleteCommandsTests.cs
tlmii: Done.

src/Microsoft.HttpRepl.IntegrationTests/Mocks/MockedFileSystem.cs
tlmii: Alright, I was able to remove it, though I'm not super thrilled with the replacement. The original code has an awkward dance around whether the headers and body are written to the same file, different files, no files, or only one of them is written to a file... and it wrote it out to a stream regardless, even when not writing it to a file. I'll resolve this for now, since the MemoryStream-backed MockedFile is gone, but I'm guessing there'll be comments on the new code.

src/Microsoft.HttpRepl.IntegrationTests/Mocks/MockedFileSystem.cs
tlmii: This is done. So far no tests need multiple files, but its set up as a Fluent call so it can be done as you described.

src/Microsoft.HttpRepl/FileSystem/IFileSystem.cs
tlmii: I went ahead and made RealFileSystem check first. There's only two cases where this was happening: 

1. WriteAllTextToFile, used when writing to a temp file, and the directory shouldn't ever not exist in that case
2. WriteAllLinesToFile, used when writing to a file that the user explicitly requested it be written to.

I think in the second case it would be a reasonable expectation of the user that we'd automatically create the directory.

src/Microsoft.HttpRepl/Microsoft.HttpRepl.csproj
```diff
@@ -13,7 +13,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNet.WebApi.Client" Version="5.2.6" />
+    <PackageReference Include="Microsoft.AspNet.WebApi.Client" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" />
```
jimmylewis: Do we still need the DI packages?

src/Microsoft.HttpRepl/FileSystem/RealFileSystem.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public string GetTempFileName()
+        {
+            return Path.GetTempFileName();
+        }
+
+        private void VerifyDirectoryExists(string path)
+        {
+            string directoryName = Path.GetDirectoryName(path);
+            if (!Directory.Exists(directoryName))
```
jimmylewis: nit: It's safe to call Directory.Create if the directory already exists, it'll noop.

tlmii: Gotcha. Simplified in c6d4a44

src/Microsoft.HttpRepl/Microsoft.HttpRepl.csproj
tlmii: Nope. Good catch. Removed in c6d4a44

src/Microsoft.HttpRepl.IntegrationTests/Commands/HttpCommandsFixture.cs
```diff
@@ -8,7 +8,7 @@ public class HttpCommandsFixture<T> : IDisposable where T : SampleApiServerConfi
         private readonly SampleApiServer _testWebServer;
 
         public T Config { get; } = new T();
-
+
```
jimmylewis: nit: cleanup trailing whitespace

PR #33, opened 2019-06-26
Added tests for SetSwaggerCommand

bhsubra:
Added tests for SetSwaggerCommand


src/Microsoft.HttpRepl.IntegrationTests/Commands/SetSwaggerCommandTests.cs
```diff
@@ -0,0 +1,324 @@
...
+using Microsoft.Repl.Parsing;
+using Moq;
+using Xunit;
+
+namespace Microsoft.HttpRepl.IntegrationTests.Commands
+{
+    public class SetSwaggerCommandTests
+    {
+        [Fact]
+        public async Task ExecuteAsync_WithExactlyOneParseResultSection_WritesToConsoleManagerError()
```
tlmii: I've been going back and forth about whether we really should be testing for this. For the end user, there's no way for them to execute the SwaggerCommand without having at least two sections in the command line - "set" and "swagger". And this test wouldn't tell us if someone made a shortcut (set-swagger?) elsewhere in the code and changed ExecuteAsync to look for it and expected it to run, because this is only testing for one section with the text "section1". 
I don't know, something feels off. Thoughts?

src/Microsoft.HttpRepl.IntegrationTests/Commands/SetSwaggerCommandTests.cs
```diff
@@ -0,0 +1,324 @@
...
+            await setSwaggerCommand.ExecuteAsync(shellState, httpState, parseResult, CancellationToken.None);
+
+            VerifyErrorMessageWasWrittenToConsoleManagerError(shellState);
+        }
+
+        [Fact]
+        public async Task ExecuteAsync_WithExactlyTwoParseResultSections_SetsHttpStateSwaggerStructureToNull()
+        {
+            MockedShellState shellState = new MockedShellState();
+            ICoreParseResult parseResult = CoreParseResultHelper.Create("section1 sections2");
```
tlmii: Similar to the comment above, it feels odd to test this with the two sections of "section1" and "sections2" when the only way the program should get here is with "set" and "swagger" as the two sections. So I don't get too annoying with comments, this applies to all the tests here that make use of a ParseResult.

src/Microsoft.HttpRepl.IntegrationTests/Commands/SetSwaggerCommandTests.cs
bhsubra: I'm testing this line of code: https://github.com/aspnet/HttpRepl/blob/dev/bhsubra/SetSwaggerCommandTests/src/Microsoft.HttpRepl/Commands/SetSwaggerCommand.cs#L246

If someone made a shortcut elsewhere in code and changed ExecuteAsync to look for it, this test needs to be updated to accommodate that change, no? Let me know if I am missing something.

src/Microsoft.HttpRepl.IntegrationTests/Commands/SetSwaggerCommandTests.cs
bhsubra: Same as above comment.

src/Microsoft.HttpRepl.IntegrationTests/Commands/SetSwaggerCommandTests.cs
bhsubra: Hmmm, now when I think of it, the error checks in ExecuteAsync(..): https://github.com/aspnet/HttpRepl/blob/dev/bhsubra/SetSwaggerCommandTests/src/Microsoft.HttpRepl/Commands/SetSwaggerCommand.cs#L246 are probably meaningless. CanHandle(..) will fail if set swagger is not present. So it won't hit ExecuteAsync(..). The check for ValidUri seems ok. We probably don't need parseResult.Sections.Count != 3 || string.IsNullOrEmpty(parseResult.Sections[2] check at all.

tlmii: I think you're now having the same back and forth I was :) Your second comment is what I was thinking... those checks shouldn't ever fail in the (current) normal flow of code. In some future change I guess there could be some other way of executing the functions that didn't call that CanHandle check first?

bhsubra: This is the current behavior:
1. Executing - set swagger does nothing. ExecuteAsync(..) gets called but since uri is not provided, it just sets the HttpState SwaggerStructure to null
2. Specifying invalid uri - set swagger test, writes an error to console. Below is complete output from console:
https://localhost:44369/~ set swagger test
Must specify a swagger document
3. When you specify sections greater than 3, says- set swagger https://localhost:44369/swagger/v1/swagger.json test2, has same console output as 2

I started working on tests for UI command and I see same weirdness in error reporting there as well. I created a new work item in azure devops and tagged Ahmed(Can't tag him here for some reason) for his input. Let me know if anyone has any concerns or anything to add here. Thanks!

src/Microsoft.HttpRepl.IntegrationTests/Commands/SetSwaggerCommandTests.cs
```diff
@@ -0,0 +1,266 @@
...
+
+            IDirectoryStructure childDirectoryStructure = directoryStructure.GetChildDirectory(expectedDirectoryName);
+            List<string> childDirectoryNames = childDirectoryStructure.DirectoryNames.ToList();
+
+            Assert.Equal(2, childDirectoryNames.Count);
+            Assert.Equal("Employees", childDirectoryNames.First());
+            Assert.Equal("Values", childDirectoryNames.ElementAt(1));
+        }
+
+        private async Task<IDirectoryStructure> GetDirectoryStructure(MockedShellState shellState, string response, string parseResultSections)
```
bhsubra: FYI, this will get moved to ICommandTestHelper once these changes are merged and when I send out pr for SetBaseCommandTests

src/Microsoft.HttpRepl/Commands/UICommand.cs
```diff
@@ -17,6 +16,12 @@ namespace Microsoft.HttpRepl.Commands
     public class UICommand : ICommand<HttpState, ICoreParseResult>
     {
         private static readonly string Name = "ui";
+        private IUriLauncher UriLauncher;
+
+        public UICommand(IUriLauncher uriLauncher = null)
+        {
+            UriLauncher = uriLauncher ?? new UriLauncher();
```
jimmylewis: Don't allow null in the constructor.  Force the dependency to be provided.

src/Microsoft.HttpRepl/HttpState.cs
```diff
@@ -69,9 +69,9 @@ public string UserProfileDir
 
         public Uri SwaggerEndpoint { get; set; }
 
-        public HttpState()
+        public HttpState(HttpClient httpClient = null)
         {
-            Client = new HttpClient();
+            Client = (httpClient == null)? new HttpClient() : httpClient;
```
jimmylewis: The constructor should force the dependency to be provided.

src/Microsoft.HttpRepl/UriLauncher.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Threading.Tasks;
+using Microsoft.HttpRepl.Resources;
+
+namespace Microsoft.HttpRepl
+{
+    internal class UriLauncher : IUriLauncher
+    {
+        public Task LaunchUriAsync(Uri uri)
+        {
+            string agent = "cmd";
```
jimmylewis: This block reads a little weird (or very platform-biased) to me.  I think it would be more clear to break it down like

```
if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
{ /* ... */ }
else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
{ /* ... */ }
else // Linux
{ /* ... */ }
```

src/Microsoft.HttpRepl.IntegrationTests/Commands/ICommandTestHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+using Microsoft.HttpRepl.IntegrationTests.Mocks;
+using Microsoft.Repl;
+using Microsoft.Repl.Commanding;
+using Microsoft.Repl.ConsoleHandling;
+using Microsoft.Repl.Parsing;
+using Moq;
+
+namespace Microsoft.HttpRepl.IntegrationTests.Commands
+{
+    public class ICommandTestHelper<T>
```
jimmylewis: Class names should not be `I*`, that's for interfaces.

src/Microsoft.HttpRepl/UriLauncher.cs
bhsubra: Updated here:https://github.com/aspnet/HttpRepl/commit/a2823f1431f0069d0fc44df13bb8b9002a08198b
and here: https://github.com/aspnet/HttpRepl/commit/21d8173d634b7464bc65ebfffda0b44d4e720fb7

src/Microsoft.HttpRepl/Commands/UICommand.cs
bhsubra: Updated here:
https://github.com/aspnet/HttpRepl/commit/dc0c5d83e599b9d8462fd595ae2e8e8e35d527cc

src/Microsoft.HttpRepl/HttpState.cs
bhsubra: Updated here:
https://github.com/aspnet/HttpRepl/commit/637d46558c13a50ae6a569c01c0dd404e16c42a6

src/Microsoft.HttpRepl.IntegrationTests/Commands/UICommandTests.cs
```diff
@@ -0,0 +1,125 @@
...
+        [Fact]
+        public void CanHandle_WithInvalidName_ReturnsNull()
+        {
+            bool? result = CanHandle(parseResultSections: "test");
+
+            Assert.Null(result);
+        }
+
+        [Fact]
+        public void CanHandle_WithValidName_ReturnsNull()
```
jimmylewis: ReturnsTrue?

src/Microsoft.HttpRepl.IntegrationTests/Commands/UICommandTests.cs
```diff
@@ -0,0 +1,125 @@
...
+            HttpState httpState = new HttpState(httpClient);
+            Uri uri = new Uri("https://localhost:44366/");
+            httpState.BaseAddress = uri;
+
+            MockUriLauncher mockUriLauncher = new MockUriLauncher(true);
+            UICommand uiCommand = new UICommand(mockUriLauncher);
+
+            var exception = await Record.ExceptionAsync(async () => await uiCommand.ExecuteAsync(shellState, httpState, parseResult, CancellationToken.None));
+
+            Assert.Null(exception);
```
jimmylewis: If you're not inspecting the exception, why not use Assert.DoesNotThrowAsync() instead of Record.ExceptionAsync()?

src/Microsoft.HttpRepl/Commands/UICommand.cs
```diff
@@ -17,6 +16,17 @@ namespace Microsoft.HttpRepl.Commands
     public class UICommand : ICommand<HttpState, ICoreParseResult>
     {
         private static readonly string Name = "ui";
+        private IUriLauncher UriLauncher;
+
+        public UICommand(IUriLauncher uriLauncher)
+        {
+            if (uriLauncher == null)
```
jimmylewis: you can one-line this, which is pretty common style now:
`UriLauncher = uriLauncher ?? throw new ArgumentNullException(nameof(uriLauncher));`

src/Microsoft.HttpRepl/HttpState.cs
```diff
@@ -69,9 +69,14 @@ public string UserProfileDir
...
-        public HttpState()
+        public HttpState(HttpClient httpClient)
         {
-            Client = new HttpClient();
+            if (httpClient == null)
+            {
+                throw new ArgumentNullException(nameof(httpClient));
+            }
+
+            Client = httpClient;
```
jimmylewis: as above, the null check can be done in one line

src/Microsoft.HttpRepl.IntegrationTests/Mocks/MockHttpMessageHandler.cs
```diff
@@ -0,0 +1,24 @@
...
+        private HttpResponseMessage response;
+
+        public MockHttpMessageHandler(HttpResponseMessage response)
+        {
+            this.response = response;
+        }
+
+        protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            var responseTask = new TaskCompletionSource<HttpResponseMessage>();
```
jimmylewis: how is this body different from `Task.FromResult(response)`?

src/Microsoft.HttpRepl.IntegrationTests/Mocks/MockHttpMessageHandler.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.HttpRepl.IntegrationTests.Mocks
+{
+    public class MockHttpMessageHandler : HttpMessageHandler
+    {
+        private HttpResponseMessage response;
+
+        public MockHttpMessageHandler(HttpResponseMessage response)
```
jimmylewis: A more flexible mock (which can be done in the future) would be to make this handler allow you to register a URL (or HttpRequestMessage if you want to allow separate HttpMethods, etc) with a response.  E.g. something like

```
var mockHttpMessageHandler = new MockHttpMessageHandler().Register("url", cannedResponse)
                                                         .Register("url2", alternateResponse);
new HttpClient(mockHttpMessageHandler);
```

src/Microsoft.HttpRepl.IntegrationTests/Commands/CommandHelper.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            return _command.GetHelpSummary(shellState, httpState);
+        }
+
+        protected async Task ExecuteAsyncWithInvalidParseResultSections(string parseResultSections, IShellState shellState, string baseAddress = null)
+        {
+            ICoreParseResult parseResult = CoreParseResultHelper.Create(parseResultSections);
+            HttpClient httpClient = new HttpClient();
+            HttpState httpState = new HttpState(httpClient);
+            httpState.BaseAddress = null;
```
jimmylewis: should this be using the passed in `baseAddress`?

src/Microsoft.HttpRepl.IntegrationTests/Commands/CommandHelper.cs
```diff
@@ -0,0 +1,79 @@
+using System.Collections.Generic;
```
jimmylewis: nit: add copyright comments at the top of any new files (or old ones that are missing them)

src/Microsoft.HttpRepl.IntegrationTests/Commands/UICommandTests.cs
bhsubra: Updated here:
https://github.com/aspnet/HttpRepl/commit/19c827eedec58a163e72966d85ba79ed6afe39da

src/Microsoft.HttpRepl/HttpState.cs
bhsubra: Updated here:
https://github.com/aspnet/HttpRepl/commit/ff18a5d8cac490242498c8397f068261b1eb7966

src/Microsoft.HttpRepl/Commands/UICommand.cs
bhsubra: Updated here:
https://github.com/aspnet/HttpRepl/commit/ff18a5d8cac490242498c8397f068261b1eb7966

src/Microsoft.HttpRepl.IntegrationTests/Mocks/MockHttpMessageHandler.cs
bhsubra: Nope, it's the same. Updated here:
https://github.com/aspnet/HttpRepl/commit/383ed286ac8470d2a34f7a4f9d2d36857aec4685

src/Microsoft.HttpRepl.IntegrationTests/Commands/CommandHelper.cs
bhsubra: Looks like I wasn't using it. Removed it here:
https://github.com/aspnet/HttpRepl/commit/5d2b32babebec634005ad2579b9e934479746316

We can add it back if we need it in some other tests.

src/Microsoft.HttpRepl.IntegrationTests/Commands/UICommandTests.cs
bhsubra: Hmmm, I don't think the latest version of xunit.assert, version=2.4.1+ has Assert.DoesNotThrowAsync. So leaving this as is. Let me know if I am missing something.

tlmii: Interesting thread on the subject (with alternatives, which includes the Record.Exception option): https://github.com/xunit/xunit/issues/188

src/Microsoft.HttpRepl.IntegrationTests/Commands/CommandHelper.cs
bhsubra: Added copyright comments to files in test projects here:
https://github.com/aspnet/HttpRepl/commit/eb0788c03704abafd1a98436677c081881d7dbec

src/Microsoft.HttpRepl.IntegrationTests/Mocks/MockHttpMessageHandler.cs
bhsubra: Sure, thanks!

I don't see a use case for the current set of tests in this pr. Let me know if I am missing something. The one you suggested might be super useful for RunCommand. Let me know if you had something else in mind.

bhsubra: What do you guys think about doing something like this?

https://github.com/xunit/xunit/issues/188#issuecomment-480822091

src/Microsoft.HttpRepl.IntegrationTests/Commands/UICommandTests.cs
bhsubra: What do you guys think about doing something like this?

xunit/xunit#188 (comment)

jimmylewis: Not sure which comment you're referring to (since the link only goes to the issue), but from reading that thread, I'm fine with the Record.Exception pattern.  I agree that Assert.Throws/Assert.DoesNotThrow breaks AAA, and I can appreciate keeping AAA more consistent to improve overall test readability.

Brad does raise a good point though about DoesNotThrow / `Assert.IsNull(exception)`.  This is sort of implicit in every test - that the Act doesn't throw (or else the test would fail anyway).  So it raises a different question - what this test trying to validate?  If you're just validating that it doesn't throw, then a passing implementation could be
```
ExecuteAsync( /* ... */ )
{
    if(IsValid(httpState))
    {
        // do nothing - yay it passes the test
    }
}
// or alternatively
ExecuteAsync( /* ... */ )
{
    if(IsValid(httpState))
    {
        try {
            uriLauncher.Launch( /* ... */ );
        } catch {}
    }
}
// or even further still, this would pass all tests in the suite:
ExecuteAsync( /* ... */ )
{
    if(IsValid(httpState))
    {
        try {
            // do dangerous stuff and mess up the state
        } catch {}
        uriLauncher.Launch( /* ... */ );
    }
}

```

This test as-implemented does not distinguish between these.  Is there something else measurable that can be validated here?

bhsubra: Updated the tests here:
https://github.com/aspnet/HttpRepl/commit/1de1b5e283e3554d2e81838bc79195767f3eed4e

The tests verify:
- LaunUriAsync(..) was called
- Exception was recorded if LaunchUriAsync(..) threw exception
- Error was written to console if base address was null

src/Microsoft.HttpRepl.IntegrationTests/Commands/SetSwaggerCommandTests.cs
```diff
@@ -0,0 +1,269 @@
...
+    {
+        public SetSwaggerCommandTests()
+          : base(new SetSwaggerCommand())
+        {
+        }
+
+        [Fact]
+        public void GetHelpDetails_WithInvalidParseResultSection_ReturnsNull()
+        {
+            string result = GetHelpDetails(parseResultSections: "section1");
```
jimmylewis: I'm really disliking this CommandHelper style, as it completely breaks AAA (Arrange/Act/Assert).  From reading these tests, I don't know what has been arranged (1) and I don't know what action is being taken (2).  In order to understand this test, I have to jump to another body of code, which takes me out of the context of what's being tested.

For point 1:
- In a positive test (i.e. happy path), I'm willing to assume a sane test default, and that the details don't matter.
- In a negative test (e.g. given an invalid input, assert error handling), I'd really like to see in the test body what the special-case arrangement is for this scenario.  It needs to be very clear what makes the input invalid.

Point 2 is actually the more important one.  Even if I assume the arrangement is correct in all cases (positive/negative/edge), I'm not actually sure what's being tested here without jumping through sources.  I'm more willing to accept helpers for arranging, but I prefer that the Act part of a test be clear and explicit - what is being called, and with what paramemters, because it helps me assess the asserts better.

To make this more readable, I'd like to see something like this:

        [Fact]
        public void GetHelpDetails_WithFirstParseResultSectionNotEqualToName_ReturnsNull()
        {
            ArrangeGetHelpDetailsInputs(commandText: "section1", out ShellState s, out HttpState h, out ICoreParseResult p);
            var subject = new SetSwaggerCommand(); // clearly define the subject being tested

            string result = subject.GetHelpDetails(s, h, p); // clearly define the action being tested

            Assert.Null(result);
        }

bhsubra: As discussed offline, will send out another pr for this change.

PR #41, opened 2019-07-02
Refactoring the Preferences code and adding tests for the PrefCommand and PreferencesProvider

tlmii:


src/Microsoft.HttpRepl/Preferences/IPreferencesProvider.cs
```diff
@@ -0,0 +1,12 @@
+using System.Collections.Generic;
+
+namespace Microsoft.HttpRepl.Preferences
+{
+    public interface IPreferencesProvider
```
tlmii: Not a fan of the name here (or of the implementing class) so open to suggestions.

src/Microsoft.HttpRepl.IntegrationTests/Mocks/MockedShellState.cs
```diff
@@ -10,31 +10,33 @@ namespace Microsoft.HttpRepl.IntegrationTests.Mocks
 {
     internal class MockedShellState : IShellState
```
tlmii: @bhsubra I refactored this so that less of it was mocked. I don't know if this will break any of your current PRs, but it works with everything that's currently in master.

src/Microsoft.HttpRepl/Microsoft.HttpRepl.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
tlmii: I'm going to guess this is some kind of Encoding issue. I thought I had undone the "changes" VS noticed here but I guess they came back before I committed.

src/Microsoft.HttpRepl.IntegrationTests/Microsoft.HttpRepl.IntegrationTests.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
tlmii: Same comment as below on the other .csproj file...

src/Microsoft.HttpRepl/UserProfile/IUserProfileDirectoryProvider.cs
```diff
@@ -0,0 +1,7 @@
+namespace Microsoft.HttpRepl.UserProfile
+{
+    public interface IUserProfileDirectoryProvider
```
tlmii: Also not a fan of this name (or the implementing class). It IS descriptive at least...

src/Microsoft.HttpRepl.IntegrationTests/Mocks/MockedShellState.cs
bhsubra: I verified it locally. This won't break any of the changes that I have in pr or locally. Thanks!

src/Microsoft.HttpRepl.IntegrationTests/Commands/PrefCommandTests.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.HttpRepl.UserProfile;
+using Microsoft.Repl.Parsing;
+using Xunit;
+
+namespace Microsoft.HttpRepl.IntegrationTests.Commands
+{
+    public class PrefCommandTests
+    {
+        [Fact]
+        public async Task ExecuteAsync_UppercaseGetCommand_CorrectOutput()
```
bhsubra: Can we add a tests for:
- Code path that hits SetSetting(..) and verify http preferences 
- SetSetting(..) with prevValue null or empty
- SetSetting(..) with prefValue
- Verify SetSetting(..) with httpState empty Preferences - writes error to console

src/Microsoft.HttpRepl.IntegrationTests/Commands/PrefCommandTests.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.HttpRepl.Commands;
+using Microsoft.HttpRepl.IntegrationTests.Mocks;
+using Microsoft.HttpRepl.Preferences;
+using Microsoft.HttpRepl.UserProfile;
+using Microsoft.Repl.Parsing;
+using Xunit;
+
+namespace Microsoft.HttpRepl.IntegrationTests.Commands
+{
+    public class PrefCommandTests
```
bhsubra: Should we also test CanHandle(..), GetHelpSummary(..), GetHelpDetails(..)? I added tests and helper here-https://github.com/aspnet/HttpRepl/pull/33/files. Let me know if you can use CommandHelper from there, I'll send out pr with just that file in that case.

src/Microsoft.HttpRepl/UserProfile/UserProfileDirectoryProvider.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.HttpRepl.UserProfile
+{
+    public class UserProfileDirectoryProvider : IUserProfileDirectoryProvider
+    {
+        public string GetUserProfileDirectory()
+        {
+            bool isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
```
bhsubra: Do we need to handle OSPlatform.OSX and OSPlatform.Linux here?

src/Microsoft.HttpRepl/Program.cs
```diff
@@ -18,7 +20,9 @@ class Program
         static async Task Main(string[] args)
         {
             IFileSystem fileSystem = new RealFileSystem();
-            HttpState state = new HttpState(fileSystem);
+            IUserProfileDirectoryProvider userProfileDirectoryProvider = new UserProfileDirectoryProvider();
+            IPreferencesProvider preferencesProvider = new PreferencesProvider(fileSystem, userProfileDirectoryProvider);
+            HttpState state = new HttpState(fileSystem, preferencesProvider);
```
bhsubra: nit: Can we create GetHttpState(..) and move everything above there?

src/Microsoft.HttpRepl/UserProfile/UserProfileDirectoryProvider.cs
tlmii: They both act the same way (using the HOME environment variable). Windows is the only weirdo of the supported platforms at the moment, as far as I can tell.

src/Microsoft.HttpRepl/Program.cs
tlmii: I moved everything except the IFileSystem into a function, since that's used elsewhere in this function.

src/Microsoft.HttpRepl/Preferences/IPreferencesProvider.cs
jimmylewis: How do you like just calling this IPreferences?

src/Microsoft.HttpRepl/Preferences/IPreferencesProvider.cs
```diff
@@ -0,0 +1,12 @@
...
+    {
+        string PreferencesFilePath { get; }
```
jimmylewis: Does this have to be exposed?  Seems like it's revealing the backing store, which should be an implementation detail (e.g. I should be able to swap in a non-file based implementation).

src/Microsoft.HttpRepl/Preferences/IPreferencesProvider.cs
```diff
@@ -0,0 +1,12 @@
...
+        IReadOnlyDictionary<string, string> GetDefaultPreferences();
```
jimmylewis: Do the defaults vary by implementation?

src/Microsoft.HttpRepl.IntegrationTests/Commands/PrefCommandTests.cs
tlmii: I added some more tests for both Get and Set. Let me know if you think I missed anything.

src/Microsoft.HttpRepl.IntegrationTests/Commands/PrefCommandTests.cs
tlmii: I added tests for CanHandle and a few for GetHelpDetails. Skipped GetHelpSummary as discussed offline.

src/Microsoft.HttpRepl/Preferences/IPreferencesProvider.cs
tlmii: Nope, doesn't need to be exposed. Removed.

src/Microsoft.HttpRepl/Preferences/IPreferencesProvider.cs
tlmii: I certainly like that more, though calling the default implementation of it Preferences leads to lots of fun namespace/class collision issues. Not sure I like the burden that added.

src/Microsoft.HttpRepl/Preferences/IPreferencesProvider.cs
tlmii: We discussed offline that no, they don't. I've modified the implementation to move it back out to `HttpState` for now. I want to take a look at what we discussed about changing the purpose of the `IPreferences` interface - from just persistence to fully controlling the preference values. But after looking into it, that "third level" of defaults we talked about is kind of spread around between `HttpState`, `RequestOrResponseConfig`, `RequestConfig`, `ResponseConfig`, `JsonConfig`. So I think we should look at changing that in a separate PR after this one is in since this one has already refactored so much.
Thoughts?

src/Microsoft.HttpRepl.IntegrationTests/Commands/PrefCommandTests.cs
```diff
@@ -0,0 +1,217 @@
...
+            Arrange(commandText, out HttpState httpState, out MockedShellState shellState, out ICoreParseResult parseResult, out PrefCommand command);
+
+            bool? canHandle = command.CanHandle(shellState, httpState, parseResult);
+
+            Assert.False(canHandle);
+            Assert.Equal(Resources.Strings.PrefCommand_Error_NoPreferenceName, shellState.ErrorMessage);
+        }
+
+        [Fact]
+        public void CanHandle_SetWithBlankName_DisplaysError()
```
jimmylewis: is the `parseResult` different between this test and the one above?  i.e. does the trailing space make a difference in the `parseResult`, or are these essentially the same tests by the time `CanHandle` is invoked?

src/Microsoft.HttpRepl.IntegrationTests/Preferences/PreferencesTests.cs
```diff
@@ -0,0 +1,97 @@
...
+This third line is invalid as well";
+            fileSystem.AddFile(preferences.PreferencesFilePath, prefsFileContent);
+
+            Dictionary<string, string> preferencesDictionary = preferences.ReadPreferences(HttpState.CreateDefaultPreferences());
+
+            Assert.Equal(expectedValue, preferencesDictionary[settingName]);
+        }
+
+        [Fact]
+        public void WritePreferences_SomeDefault_OnlyWritesNonDefaultValues()
```
jimmylewis: it feels like there's a second test lingering in the OnlyWritesNonDefaultValues shadow:

1. load prefs from a file that contains a non-default setting
2. Set that pref to the default value
3. Write preferences
4. Verify that preference is removed from the file?

src/Microsoft.HttpRepl.IntegrationTests/Commands/PrefCommandTests.cs
```diff
@@ -0,0 +1,217 @@
+using System;
```
jimmylewis: nit: add copyright comments to the top of any new files (or files which didn't have it already)

src/Microsoft.HttpRepl/Preferences/Preferences.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            if (_fileSystem.FileExists(PreferencesFilePath))
+            {
+                string[] prefsFile = _fileSystem.ReadAllLinesFromFile(PreferencesFilePath);
+
+                foreach (string line in prefsFile)
+                {
+                    int equalsIndex = line.IndexOf('=');
+
+                    if (equalsIndex < 0)
```
jimmylewis: is `=foo` a valid line?  equalsIndex would be 0, so I'd get an empty name?

src/Microsoft.HttpRepl.IntegrationTests/Commands/PrefCommandTests.cs
tlmii: The parseResult is different in this case. In the SetWithNoName case, parseResult only has two sections (pref and set), so the check for # of arguments causes it to return false. In this case, parseResult has four sections (pref, set and two blank strings), so the check for whether or not the third argument is blank causes it to return false.

src/Microsoft.HttpRepl.IntegrationTests/Commands/PrefCommandTests.cs
tlmii: Saw your comments on Bhavya's PR so the commit just above this comment (79435d7) added them to the new files in this PR.

src/Microsoft.HttpRepl/Preferences/IPreferencesProvider.cs
jimmylewis: yeah, I see what you mean about the name.  The default could be something like UserFolderPreferences or something to indicate what's unique to that implementation.

src/Microsoft.HttpRepl/Preferences/Preferences.cs
tlmii: It wouldn't fail, but it wouldn't be useful. So I changed it to actively skip lines like that.

src/Microsoft.HttpRepl.IntegrationTests/Preferences/PreferencesTests.cs
tlmii: Good call. I added a test for that.

src/Microsoft.HttpRepl/Preferences/IPreferencesProvider.cs
tlmii: Works for me. Done.

PR #45, opened 2019-07-03
Add a basic integration test for a Swagger-backed endpoint

jimmylewis:
The only functionality that is mocked out in this new test is the ConsoleManager, because the default implementation doesn't work well when executing via tests.  Other than that, this is running everything fully integrated against a live endpoint.


build/PackageVersions.targets
```diff
@@ -7,6 +7,7 @@
   <ItemGroup>
     <PackageReference Update="Microsoft.AspNet.WebApi.Client" Version="5.2.6" />
     <PackageReference Update="Moq" Version="4.10.1" PrivateAssets="All" />
-    <PackageReference Update="Newtonsoft.Json" Version="11.0.2" PrivateAssets="All" />
+    <PackageReference Update="Newtonsoft.Json" Version="12.0.1" PrivateAssets="All" />
```
tlmii: Just out of curiosity, what required the upgrade in Json.Net? Swashbuckle?

src/Microsoft.HttpRepl.IntegrationTests/Mocks/LoggingConsoleManagerDecorator.cs
```diff
@@ -0,0 +1,89 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Text;
+using System.Threading;
+using Microsoft.Repl.ConsoleHandling;
+
+namespace Microsoft.HttpRepl.IntegrationTests.Mocks
+{
+    class LoggingConsoleManagerDecorator : IConsoleManager
```
tlmii: nit: access modifier (class and fields)

src/Microsoft.HttpRepl.IntegrationTests/SampleApi/SampleApiServer.cs
```diff
@@ -1,57 +1,69 @@
...
+                               app.UseRouter(routes);
 
-        public void Stop()
-        {
-            _Host.StopAsync();
-        }
+                               app.UseSwagger();
+                           })
+                           .Build();
+            }
```
tlmii: nit: I think the indentation from here down is off

src/Microsoft.Repl/Shell.cs
```diff
@@ -17,8 +18,8 @@ public Shell(IShellState shellState)
             ShellState = shellState;
         }
 
-        public Shell(ICommandDispatcher dispatcher, ISuggestionManager suggestionManager = null)
-            : this(new ShellState(dispatcher, suggestionManager))
+        public Shell(ICommandDispatcher dispatcher, ISuggestionManager suggestionManager = null, IConsoleManager consoleManager = null)
+            : this(new ShellState(dispatcher, suggestionManager, consoleManager: consoleManager))
```
tlmii: This is probably a change for a different PR, but should we consider changing this (the ShellState constructor, really) so it requires the constructor parameters and doesn't create defaults if they're null?

build/PackageVersions.targets
jimmylewis: yup.

src/Microsoft.Repl/Shell.cs
jimmylewis: I think this gets into having a consolidated composition root (really the Shell constructor should not be new'ing ShellState either), and I'm gonna call out-of-scope for this PR.

PR #48, opened 2019-07-05
Removing ICommand.GetHelpSummary and all of its implementations

### LINKED ISSUE #43: ICommand.GetHelpSummary is never called
The abstract definition is here: https://github.com/aspnet/HttpRepl/blob/master/src/Microsoft.Repl/Commanding/ICommand.cs and it is overridden in the various derived classes, but it is never called.

We should remove it (or, if this is a mistake somehow, correct the mistake).

tlmii:
Fixes #43 - ICommand.GetHelpSummary, and all of its implementations are never called. In many cases, it's just a duplicate of GetHelpDetails anyway.

### DIFF
--- a/src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
+++ b/src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
@@ -498,11 +498,6 @@ namespace Microsoft.HttpRepl.Commands
             return helpText.ToString();
         }
 
-        public override string GetHelpSummary(IShellState shellState, HttpState programState)
-        {
-            return $"{Verb.ToLowerInvariant()} - Issues a {Verb.ToUpperInvariant()} request";
-        }
-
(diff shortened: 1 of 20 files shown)


jimmylewis:
It seems like these ought to have been used by `HelpCommand.CoreGetHelp()`, which currently has a hardcoded list of commands and **one-line** help summaries.  Are we at a crossroads of GetCoreHelp being static text vs. dynamically generated from the corresponding `ICommand` implementation?

tlmii:
Yeah, the static vs dynamic thing makes this awkward. The help text in `HelpCommand.CoreGetHelp` seems to be sort of "curated" - commands in a specific order, leaving out some commands (`config` and `set diag` at least), etc. I guess we could new up those commands in `HelpCommand.CoreGetHelp` (or pull them out of the dispatcher by type) and call GetHelpSummary from them so at least we don't have the literals in there.

jimmylewis:
Even with a curated list, it seems like GetCoreHelp should be leveraging each command's GetHelpSummary() - i.e. GetCoreHelp() shouldn't know the help summaries of each command.

tlmii:
Closing this PR as we're changed directions and gone with #62 instead.

PR #49, opened 2019-07-05
Fixes SetHeader command to not blindly return documentation on GetHelpDetails

### LINKED ISSUE #34: SetHeaderCommand help is displayed instead of SetSwaggerCommand help for "help set swagger"
It looks like SetHeaderCommand is just coded to return the help regardless of whether the input to it is valid. So when the help command iterates through the commands looking for the first one that is valid, SetHeaderCommand returns a value when it shouldn't.

tlmii:
Fixes #34 - GetHelpDetails is used in a way that requires the callee to make sure it should actually respond. SetHeaderCommand did not do that, so it responded when the help request was actually for a command that came later in the list - SetSwaggerCommand, for instance.

### DIFF
--- a/src/Microsoft.HttpRepl/Commands/SetHeaderCommand.cs
+++ b/src/Microsoft.HttpRepl/Commands/SetHeaderCommand.cs
@@ -45,12 +45,17 @@ namespace Microsoft.HttpRepl.Commands
 
         public string GetHelpDetails(IShellState shellState, HttpState programState, ICoreParseResult parseResult)
         {
-            var helpText = new StringBuilder();
-            helpText.Append("Usage: ".Bold());
-            helpText.AppendLine("set header {name} [value]");
-            helpText.AppendLine();
-            helpText.AppendLine("Sets or clears a header. When [value] is empty the header is cleared.");
-            return Description;
+            if (parseResult.Sections.Count > 1 && string.Equals(parseResult.Sections[0], Name, StringComparison.Ordinal) && string.Equals(parseResult.Sections[1], SubCommand, StringComparison.OrdinalIgnoreCase))
+            {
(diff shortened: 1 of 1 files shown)


src/Microsoft.HttpRepl/Commands/SetHeaderCommand.cs
```diff
@@ -45,12 +45,17 @@ public Task ExecuteAsync(IShellState shellState, HttpState programState, ICorePa
 
         public string GetHelpDetails(IShellState shellState, HttpState programState, ICoreParseResult parseResult)
         {
-            var helpText = new StringBuilder();
-            helpText.Append("Usage: ".Bold());
-            helpText.AppendLine("set header {name} [value]");
-            helpText.AppendLine();
-            helpText.AppendLine("Sets or clears a header. When [value] is empty the header is cleared.");
-            return Description;
+            if (parseResult.Sections.Count > 1 && string.Equals(parseResult.Sections[0], Name, StringComparison.Ordinal) && string.Equals(parseResult.Sections[1], SubCommand, StringComparison.OrdinalIgnoreCase))
```
jimmylewis: Could we wrap this line into a common method to improve readability and reduce duplication, since it's used verbatim in at least 4 commands (and duplicated in `CanHandle` as well).

```
if(CommandParsingUtilitiesToBeGivenABetterName.ParseResultMatches(commandName: Name, subCommandName: SubCommand))
{ ... }
```

There isn't a common base to put the implementation in, so an extension method or helper may be appropriate...

tlmii: Per our discussion offline, I've made #60 to address a larger refactoring of this since it is reused often.

PR #53, opened 2019-07-05
Rework preferences

tlmii:
This PR is a response to conversations I had with @jimmylewis while working on #41 (the tests for `PrefCommand` and related classes). The preferences system was spread out across multiple classes in a way that made it hard to follow and test.

This PR attempts to correct that by moving most of the logic into the `IPreferences` implementation `UserFolderPreferences`.

In particular:

1. Preferences are no longer held in memory in the `HttpState` object and accessed directly off that in an almost-global manner. 
2. There's no longer a requirement to explicitly read the settings at startup and write them after every setting change. That is handled by the `UserFolderPreferences` under the hood, allowing the settings to be lazy loaded and, if we chose, delay-written. That latter part is not implemented as part of this PR right now.
3. The logic of handling the defaults values versus the current values is no longer handled in the `PrefCommand`, and is instead part of `UserFolderPreferences`. `PrefCommand` can thus focus on what it really needs to do, which is handle the user command and route it appropriately to the preferences system.
4. The default preferences were moved out of `HttpState` and into `Program`, since they're not really tied to `HttpState`. They are currently still used by the test cases, but that is now encapsulated a little better to allow separation if we chose.

### DIFF
--- a/src/Microsoft.HttpRepl.IntegrationTests/Preferences/UserFolderPreferencesTests.cs
+++ b/src/Microsoft.HttpRepl.IntegrationTests/Preferences/UserFolderPreferencesTests.cs
@@ -63,9 +63,13 @@ This third line is invalid as well";
 
             // Add one and change one
             // Only the changes should be written, not any of the defaults.
-            preferences.SetValue(WellKnownPreference.DefaultEditorCommand, defaultEditor);
-            preferences.SetValue(WellKnownPreference.ErrorColor, errorColor);
+            bool succeeded = preferences.SetValue(WellKnownPreference.DefaultEditorCommand, defaultEditor);
 
+            Assert.True(succeeded);
+
+            succeeded = preferences.SetValue(WellKnownPreference.ErrorColor, errorColor);
+
+            Assert.True(succeeded);
             Assert.Equal(expected, fileSystem.ReadFile(preferences.PreferencesFilePath));
         }
 
@@ -88,8 +92,9 @@ This third line is invalid as well";
 
             // Now change it to the default value, write it back to the file system and
             // validate that it was removed from the file
-            preferences.SetValue(WellKnownPreference.ProtocolColor, defaultPreferences[WellKnownPreference.ProtocolColor]);
-            
+            bool succeeded = preferences.SetValue(WellKnownPreference.ProtocolColor, defaultPreferences[WellKnownPreference.ProtocolColor]);
+
+            Assert.True(succeeded);
             Assert.Equal(string.Empty, fileSystem.ReadFile(preferences.PreferencesFilePath));
         }
 
--- a/src/Microsoft.HttpRepl.IntegrationTests/Preferences/UserFolderPreferencesTests.cs
+++ b/src/Microsoft.HttpRepl.IntegrationTests/Preferences/UserFolderPreferencesTests.cs
@@ -2,7 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
+using Microsoft.Extensions.ObjectPool;
 using Microsoft.HttpRepl.FileSystem;
 using Microsoft.HttpRepl.IntegrationTests.Mocks;
 using Microsoft.HttpRepl.Preferences;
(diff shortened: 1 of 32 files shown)


src/Microsoft.HttpRepl/Program.cs
```diff
@@ -23,9 +24,8 @@ public static async Task Main(string[] args)
         }
 
         public async Task Start(string[] args, IConsoleManager console)
-        { 
-            IFileSystem fileSystem = new RealFileSystem();
-            HttpState state = CreateHttpState(fileSystem);
+        {           
+            HttpState state = CreateHttpState(out IFileSystem fileSystem, out IPreferences preferences);
```
jimmylewis: Hm, the purist in me is saying that this is fragmenting the composition root - the dependencies are being initialized in a separate place from the rest of the dependency graph (here).  That fragmentation was actually introduced in the last Preferences change, when `CreateHttpState` was created.

What do you think about formalizing composition more?  It's really not a responsibility of the Start method - what if we had an explicit root like `ComposeDependencies(IConsole console, out Shell shell, out IPreferences prefs)` (to start with - it may evolve over time, though hopefully not - console is already kind of a hack to enable the integration tests, but the rest of the composition for the app should be internal to the composition root even if types/dependencies change), and that would take the CreateHttpState and dispatcher setup all into one place, leaving Start as something like (pseudo):

```
Start(string[] args, IConsole console) 
{
    ComposeDependencies(console, out Shell shell, out IPreferences prefs);
    if (Console.IsOutputRedirected && !console.AllowOutputRedirection)
    {
        Reporter.Error.WriteLine(Resources.Strings.Error_OutputRedirected.SetColor(prefs.GetValue(WellKnownPreference.ErrorColor)));
        return;
    }

    CancellationTokenSource source = new CancellationTokenSource();
~~    Shell shell = new Shell(dispatcher, consoleManager: console);~~ // this line no longer needed
    shell.ShellState.ConsoleManager.AddBreakHandler(() => source.Cancel());
    /* etc... */
}
```

src/Microsoft.HttpRepl/Preferences/UserFolderPreferences.cs
```diff
@@ -27,15 +46,80 @@ public string PreferencesFilePath
...
+        public UserFolderPreferences(IFileSystem fileSystem, IUserProfileDirectoryProvider userProfileDirectoryProvider, IDictionary<string, string> defaultPreferences)
         {
             _fileSystem = fileSystem;
             _userProfileDirectoryProvider = userProfileDirectoryProvider;
+            DefaultPreferences = SetupDefaults(defaultPreferences);
+        }
+
+        public AllowedColors GetColorValue(string preference, AllowedColors defaultValue = AllowedColors.None)
+        {
+            if (!Preferences.TryGetValue(preference, out string preferenceValueString) || !Enum.TryParse(preferenceValueString, true, out AllowedColors result))
```
jimmylewis: There's a lot of branches added to this class (each of these ifs has 3 branches).  How's code coverage looking?

src/Microsoft.HttpRepl/Program.cs
tlmii: Makes sense to me. I gave it a shot in c430a00, let me know what you think.

src/Microsoft.HttpRepl/Preferences/UserFolderPreferences.cs
tlmii: Note that those three functions were basically just moved from the WellKnownPreferences.cs file. That said, I had a bunch of tests in the works for them today and they just went in with c430a00.

src/Microsoft.HttpRepl.IntegrationTests/Preferences/UserFolderPreferencesTests.cs
```diff
@@ -82,37 +81,159 @@ public void WritePreferences_ChangeNonDefaultToDefault_RemovesDefaultValue()
...
+            bool succeeded = preferences.SetValue(WellKnownPreference.ProtocolColor, defaultPreferences[WellKnownPreference.ProtocolColor]);
 
             Assert.True(succeeded);
             Assert.Equal(string.Empty, fileSystem.ReadFile(preferences.PreferencesFilePath));
         }
 
+        [Fact]
+        public void GetColorValue_DoesNotExist_ReturnsDefault()
+        {
+            ConfirmColorValue(expected: AllowedColors.Magenta,
```
jimmylewis: If all these tests are really identical except for inputs, use `[Theory]` instead of `[Fact]`.  This also helps keep the test body in AAA form without having to jump elsewhere to read the code.

src/Microsoft.HttpRepl/Program.cs
jimmylewis: I like it 👍

src/Microsoft.HttpRepl.IntegrationTests/Preferences/UserFolderPreferencesTests.cs
tlmii: Done in b318acc, though I'll admit its annoying to not be able to specify a test name for each set of data.

PR #61, opened 2019-07-08
Initialize HTTP request headers Dictionary as case-insensitive

scottaddie:
This is a small fix for something I noticed while writing the doc. The generic dictionary used for storing HTTP request headers should be created as case-insensitive (as most others in the code base are). Without that change, the first condition in the following if statement is problematic:

https://github.com/aspnet/HttpRepl/blob/e261002df7f7d96f26337b2057b6e3f31dfe88da/src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs#L128

Imagine a situation in which I execute the following command:

```console
https://localhost:5001/People~ post -h Content-Type=application/json
```

The HTTP request header uses a different casing than what the if statement is expecting. With a case-insensitive dictionary, the casing in this situation no longer matters.

### DIFF
--- a/src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
+++ b/src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
@@ -99,7 +99,7 @@ namespace Microsoft.HttpRepl.Commands
                 }
             }
 
-            Dictionary<string, string> thisRequestHeaders = new Dictionary<string, string>();
+            Dictionary<string, string> thisRequestHeaders = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
 
             foreach (InputElement header in commandInput.Options[HeaderOption])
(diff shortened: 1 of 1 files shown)


tlmii:
HTTP Header names are case-insensitive, so this change makes sense to me. Adding others as reviewers just in case before I merge this.

jimmylewis:
It'd be nice to have a regression test added to verify we're handling the case-insensitivity correctly.  @tlmii do we have test coverage for this code path yet, or is that one we're still working up to?

tlmii:
@jimmylewis We don't, at the moment. Once #33 is in, we'll evaluate the best path forward to fully test `BaseHttpCommand.ExecuteAsync` and `BaseHttpCommand.HandleResponseAsync`. For now, as discussed offline, I'll go ahead and merge this change to get the fix in.

PR #62, opened 2019-07-08
Using ICommand.GetHelpSummary from the HelpCommand

### LINKED ISSUE #43: ICommand.GetHelpSummary is never called

tlmii:
This is a replacement PR for #48 to fix #43. 

`HelpCommand.GetCoreHelp(...)` had all of the help summaries hard-coded. This PR changes it to use the ICommand.GetHelpSummary(...) for each command. It also moves the strings used for GetHelpSummary into Resources.

### DIFF
--- a/src/Microsoft.HttpRepl/Commands/HelpCommand.cs
+++ b/src/Microsoft.HttpRepl/Commands/HelpCommand.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Windows.Input;
 using Microsoft.HttpRepl.Preferences;
 using Microsoft.HttpRepl.Suggestions;
 using Microsoft.Repl;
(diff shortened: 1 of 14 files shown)


src/Microsoft.HttpRepl/Commands/HelpCommand.cs
```diff
@@ -221,49 +222,56 @@ public void CoreGetHelp(IShellState shellState, ICommandDispatcher<HttpState, IC
...
+            output.AppendLine($"{"run",navCommandColumn}{dispatcher.GetCommand<RunCommand>().GetHelpSummary(shellState, programState)}");
+            output.AppendLine($"{"ui",navCommandColumn}{dispatcher.GetCommand<UICommand>().GetHelpSummary(shellState, programState)}");
             output.AppendLine();
             output.AppendLine("Use help <COMMAND> to learn more details about individual commands. e.g. `help get`".Bold().Cyan());
             output.AppendLine();
 
             shellState.ConsoleManager.Write(output.ToString());
         }
+
+        private static T GetCommand<T, TState, TParseResult>(ICommandDispatcher<TState, TParseResult> dispatcher) where T: class, ICommand where TParseResult: ICoreParseResult
```
jimmylewis: Is this used?  It looks like all the calls are using the extension method.

tlmii: Whoops. No, that was a precursor to the extension method that I forgot to remove. Removed in df3f30f.

Issue #51, opened 2019-07-05
Navigation to nonexistent endpoint is allowed


scottaddie:
If not taking advantage of tab completion, it's possible to navigate to an endpoint that doesn't exist. Consider the following web API containing `Fruits` and `People` endpoints.

```console
https://localhost:5001/~ ls
.        []
Fruits   [get|post]
People   [get|post]

https://localhost:5001/~ cd pe
/pe    []
```

When running `cd pe` in the preceding example, I would've expected:

* Navigation away from the web API root to be prevented.
* An error message such as "The pe endpoint doesn't exist".

/cc @tlmii

tlmii:
I'm not sure what the proper expectation here is. 

When you don't have a swagger endpoint set (so the system doesn't know about the directory structure), you can still move around with the cd command to build up your proper URL to call. I think not limiting the user in that case makes sense since otherwise people without a swagger endpoint wouldn't be able to really use the tool.

But when we do know what the API structure is supposed to look like, should we limit the user? I'm not sure we should. There could be other endpoints outside of the swagger definition for whatever reason.

@jodavis any thoughts? @vijayrkn any historical context to that decision?

jodavis:
@tlmii I definitely agree that in the case where you don't have a swagger endpoint set, we should not restrict the path. 

For consistency, I'm inclined to say we shouldn't be restrictive even when a swagger endpoint is set. Like you said, maybe there's some path that's not defined in swagger, and in that case I could see the user removing and re-adding the swagger definition to access the undocumented APIs.

But we might help the user by displaying a message as part of the command prompt. Something like

```
Warning: The 'pe' endpoint does not exist in the swagger definition
https://localhost:5001/~ |
```

(Replace any terminology that would be more consistent with the rest of the tool.)

That would call attention to a typo in the selected endpoint, but it could be ignored if the omitted endpoint was intentional.

tlmii:
Now that #64 has been merged, I'll go ahead and close this.

PR #64, opened 2019-07-09
Adding a warning when ChangeDirectoryCommand is used to navigate to an endpoint not listed in the current swagger metadata.

tlmii:
Addresses #51 by adding a warning, when the directory structure is set AND the current endpoint is unknown, that states that the current endpoint is not present in the swagger definition/metadata.


src/Microsoft.HttpRepl/Resources/Strings.resx
```diff
@@ -120,6 +120,10 @@
   <data name="ChangeDirectoryCommand_HelpSummary" xml:space="preserve">
     <value>Append the given directory to the currently selected path, or move up a path when using `cd ..`</value>
   </data>
+  <data name="ChangeDirectoryCommand_Warning_UnknownEndpoint" xml:space="preserve">
+    <value>Warning: The '/{0}' endpoint is not present in the swagger metadata</value>
```
scottaddie:
```suggestion
    <value>Warning: The '/{0}' endpoint is not present in the Swagger metadata</value>
```

src/Microsoft.HttpRepl/Commands/ChangeDirectoryCommand.cs
```diff
@@ -57,6 +57,11 @@ protected override Task ExecuteAsync(IShellState shellState, HttpState programSt
                 {
                     IDirectoryStructure s = programState.Structure.TraverseTo(programState.PathSections.Reverse());
 
+                    if (s.RequestInfo == null || s.RequestInfo.Methods.Count == 0)
```
jimmylewis: a) What do you think about combining this if with the ternary operator on the next line (which has the opposite predicate) into a consolidated if-else?

b) tests?

jimmylewis: also, nit: use `is null` for null checks.

tlmii: Done, plus some other cleanup in 1332027 and 2549121.

src/Microsoft.HttpRepl.Tests/HttpStateTests.cs
```diff
@@ -0,0 +1,59 @@
...
+            state.PathSections.Push("SecondDirectory");
+
+            string result = state.GetRelativePathString();
+
+            Assert.Equal(expected, result);
+        }
+
+        private static HttpState SetupHttpState()
+        {
+            IFileSystem fileSystem = new RealFileSystem();
```
jimmylewis: Shouldn't unit tests generally use a fake filesystem?

src/Microsoft.HttpRepl/Commands/ChangeDirectoryCommand.cs
```diff
@@ -57,17 +57,26 @@ protected override Task ExecuteAsync(IShellState shellState, HttpState programSt
                 {
                     IDirectoryStructure s = programState.Structure.TraverseTo(programState.PathSections.Reverse());
 
-                    if (s.RequestInfo == null || s.RequestInfo.Methods.Count == 0)
+                    string thisDirMethod;
+
+                    // If there's no RequestInfo/Methods AND there's no (sub)DirectoryNames, it must be an
+                    // autogenerated directory 
+                    if ((s.RequestInfo?.Methods is null || s.RequestInfo.Methods.Count == 0) && (s.DirectoryNames is null || !s.DirectoryNames.Any()))
```
jimmylewis: instead of a comment explaining the long predicate, how about:

```
bool hasNoRequestInfo = s.RequestInfo?.Methods is null || s.RequestInfo.Methods.Count == 0;
bool hasNoDirectoryNames = s.DirectoryNames is null || !s.DirectoryNames.Any();
if(hasNoRequestInfo && hasNoDirectoryNames)
{ /* ... */ }
```

src/Microsoft.HttpRepl/HttpState.cs
```diff
@@ -289,5 +289,17 @@ public Uri GetEffectivePath(IReadOnlyList<string> sections, bool requiresBody, o
 
             return builder.Uri;
         }
+
+        internal string GetRelativePathString()
```
jimmylewis: Why not just make this public?

IMO, methods should be either public (part of the behavior contract of the type), or private (implementation detail).  I prefer to avoid internal methods (the same rules don't apply for types).

src/Microsoft.HttpRepl.Tests/HttpStateTests.cs
tlmii: Whoops. Yeah. There wasn't a fake one in this assembly, so I created a true stub (since it's never actually called in these tests) in afc0086. Tried to follow the naming conventions we talked about yesterday.

src/Microsoft.HttpRepl/Commands/ChangeDirectoryCommand.cs
tlmii: So the comment wasn't as much about explaining the predicate as making our assumption (that the lack of those two things meant it was auto-generated) explicit rather than implicit by the code. I like the way you listed it better anyway, though, so I changed it and modified the comment a little bit in https://github.com/aspnet/HttpRepl/commit/afc0086f7d73c95c514a5fc9defe7f89d80f4a76.

src/Microsoft.HttpRepl/HttpState.cs
tlmii: That's fair. I changed it to public. My instinct tends to be "as restricted as possible to make it still work" but I see your point.

src/Microsoft.HttpRepl.Tests/Fakes/FileSystemStub.cs
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.HttpRepl.FileSystem;
+
+namespace Microsoft.HttpRepl.Tests.TestDoubles
+{
+    internal class FileSystemStub : IFileSystem
```
jimmylewis: This is kind of redundant with MockedFileSystem in the integration tests project (which, as we've discussed, probably needs to be moved back to unit tests at some point - integration tests shouldn't be using a fake).  Just pointing it out as something to cleanup when we do that merge.

src/Microsoft.HttpRepl/Properties/AssemblyInfo.cs
```diff
@@ -1,3 +1,4 @@
 using System.Runtime.CompilerServices;
 
+[assembly: InternalsVisibleTo("Microsoft.HttpRepl.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
```
jimmylewis: Do we still need this?

src/Microsoft.HttpRepl.Tests/Fakes/FileSystemStub.cs
tlmii: I edited my earlier reply after I initially posted it, so you probably missed the second half of it:  
> I imagine we might want to merge the implementations but figured I would leave that for when we get to discussing moving the current integration-tests-that-aren't into this assembly.

src/Microsoft.HttpRepl/Properties/AssemblyInfo.cs
tlmii: Nope. Removed in https://github.com/aspnet/HttpRepl/pull/64/commits/e524082ce44a718e5ce0bf009da6b5589f25ef31.

PR #65, opened 2019-07-10
Update CommandHelper and tests to follow AAA pattern

bhsubra:
Changes in this pr addresses the below comment:
https://github.com/aspnet/HttpRepl/pull/33/files#r301319287

From pr:
https://github.com/aspnet/HttpRepl/pull/33


src/Microsoft.HttpRepl.IntegrationTests/Commands/CommandHelper.cs
```diff
@@ -1,73 +1,29 @@
...
-using Microsoft.Repl.Commanding;
 using Microsoft.Repl.ConsoleHandling;
 using Microsoft.Repl.Parsing;
 using Moq;
 
 namespace Microsoft.HttpRepl.IntegrationTests.Commands
 {
-    public class CommandHelper<T>
-        where T : ICommand<HttpState, ICoreParseResult>
+    public class CommandHelper
```
tlmii: This probably could have gone into the other PR, but I didn't want to delay it. It sort of feels weird to have a class named `*Helper` that is intended to be the base class for other classes. `*Helper` to me implies static methods or something to be used by something else. What do you think about either renaming it to something like `BaseCommandTests`/`CommandTestsBase` or making the three functions static (since I don't think they depend on anything anymore). cc @jimmylewis since I'm guessing he'll have opinions.

src/Microsoft.HttpRepl.IntegrationTests/Mocks/MockedShellState.cs
```diff
@@ -11,7 +11,7 @@
 
 namespace Microsoft.HttpRepl.IntegrationTests.Mocks
 {
-    internal class MockedShellState : IShellState
+    public class MockedShellState : IShellState
```
tlmii: What was the intent here?

bhsubra:
This pr got messed up after I merged #33. I will close this one and create a new pr. I will look at existing comments before I send out a new one.

PR #69, opened 2019-07-10
Add an integration test with the Swagger endpoint disabled

jimmylewis:
This adds a flag to the test fixture to toggle whether the Swagger
endpoint is active, which allows tests to be written with it on or off.

Still to do: find a way to toggle it on each test run.  Ideally, we
would have a single Theory test which would take as inputs whether the
Swagger endpoint is active, and the corresponding expected output.
Currently the tests have to be separate because the server is started
during Fixture initialization, which is before the test method receives
control.


jimmylewis:
Yes, I'd like to see that at some point - namely, we have 3 endpoint readers implemented, it'd be nice to be able to target each of them with the "live" endpoint.  But for now I want to get the basics in so we can have some level of test authoring.

I'd also prefer to split these out differently once we sort out integration vs. unit tests.  E.g. these should be ChangeDirectory tests instead of Swagger/NonSwagger tests, but I didn't want to confuse them with the "unit"-like tests currently in this project.

PR #77, opened 2019-07-12
Added tests for - SetHeaderCommand, SetBaseCommand, RunCommand

bhsubra:
Added tests for - SetHeaderCommand, SetBaseCommand, RunCommand


src/Microsoft.HttpRepl.IntegrationTests/Commands/CommandTestsBase.cs
```diff
@@ -19,11 +19,13 @@ public class CommandTestsBase
...
-            out ICoreParseResult parseResult)
+            out ICoreParseResult parseResult,
+            int caretPosition = 0,
+            string responseContent = null)
         {
-            parseResult = CoreParseResultHelper.Create(parseResultSections);
+            parseResult = CoreParseResultHelper.Create(parseResultSections, caretPosition);
             shellState = new MockedShellState();
-            httpState = GetHttpState(string.Empty);
+            httpState = GetHttpState(responseContent);
```
tlmii: It looks like you change the default from `string.Empty` to `null` here. Just want to make sure that doesn't have any downstream implications. Guessing not, but thought I'd check.

src/Microsoft.HttpRepl.IntegrationTests/Mocks/MockDirectoryStructure.cs
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.HttpRepl.IntegrationTests.Mocks
+{
+    internal class MockDirectoryStructure : IDirectoryStructure
```
tlmii: I'm not convince the Fake is necessary here, since `DirectoryStructure` is so lightweight as it is. It does keep the calls to it in the tests cleaner, and it's not what you're testing against, so it's probably ok. Just wanted to throw it out there.

src/Microsoft.HttpRepl.IntegrationTests/Commands/CommandTestsBase.cs
bhsubra: Nope, it should have any implications. I changed it as it won't allow string.Empty as default. Added null check here just in case:
https://github.com/aspnet/HttpRepl/commit/c696814b69478ae5a0162d45379a90f5a7ea384c

src/Microsoft.HttpRepl.IntegrationTests/Mocks/MockDirectoryStructure.cs
bhsubra: That's true. Removed it here:
https://github.com/aspnet/HttpRepl/commit/7b4b20697f18184252ddd9c3a5ecd278de1d5c55

Thanks!

src/Microsoft.HttpRepl.IntegrationTests/Commands/CommandTestsBase.cs
jimmylewis: string.empty is readonly, but not const (default params must be constant values).  I believe you can use `""` as a constant non-null empty string value for parameter defaults.  Or use continue to use null, and make sure it's handled correctly everywhere.

bhsubra: "" works, updated here:
https://github.com/aspnet/HttpRepl/commit/f0d84962d91caabf832097b6957e32b1fce4c265

Thanks!

PR #83, opened 2019-07-15
Refactor ICoreParseResult section checking

### LINKED ISSUE #60: ICoreParseResult section checking needs refactoring
This stems from @jimmylewis's comment on https://github.com/aspnet/HttpRepl/pull/49#discussion_r301213844.

There are a lot of cases in the code where we have something similar to:

```
ICoreParseResult parseResult;
...
if (parseResult.Sections.Count > X && 
    parseResult.Sections[0] == "Y" && 
    parseResult.Sections[1] == "Z")
{
    ...
}
```

But they have slight variances. "Y" is usually the Name of the command, "Z" (if present) is usually the SubCommand. But X van vary independently of the presence of "Z" and the operator can sometimes be `==` instead of `>`. 

We should refactor those to make them more consistent (if possible) and more readable.

tlmii:
Fixes #60 by introducing some extension methods for ICoreParseResult.

Originally, there was a lot of code such as:

```C#
bool canHandle = parseResult.Sections.Count == 2 && string.Equals(parseResult.Sections[0], Name, StringComparison.OrdinalIgnoreCase) && string.Equals(parseResult.Sections[1], SubCommand, StringComparison.OrdinalIgnoreCase));
bool canHandle = parseResult.Sections.Count == 2 && string.Equals(parseResult.Sections[0], Name, StringComparison.OrdinalIgnoreCase) && string.Equals(parseResult.Sections[1], SubCommand, StringComparison.OrdinalIgnoreCase));
```

With this PR, those calls could be replaced with:
```C#
bool canHandle = parseResults.ContainsExactly(Name, SubCommand);
bool canHandle = parseResults.ContainsAtLeast(Name, SubCommand);
```
The simple forms of the method use the length of the arguments passed in along with a default of `StringComparison.OrdinalIgnoreCase` to find the result. There are overloads that allow you to specify those directly if necessary, however:

```C#
public static bool ContainsExactly(this ICoreParseResult parseResult, int length, StringComparison stringComparison, params string[] sections)
{
...
}
public static bool ContainsAtLeast(this ICoreParseResult parseResult, int minimumLength, StringComparison stringComparison, params string[] sections)
{
...
}
```
This is most commonly used (currently) when you want to  say something like "there must be at least two sections, but all I know is that the first one needs to be Name. So you'd do something like:

```C#
bool canHandle = parseResults.ContainsAtLeast(minimumLength: 2, Name);
```

### DIFF
--- a/src/Microsoft.Repl.Tests/Microsoft.Repl.Tests.csproj
+++ b/src/Microsoft.Repl.Tests/Microsoft.Repl.Tests.csproj
@@ -12,8 +12,4 @@
     <ProjectReference Include="..\Microsoft.Repl\Microsoft.Repl.csproj" />
   </ItemGroup>
 
-  <ItemGroup>
-    <Folder Include="Parsing\" />
-  </ItemGroup>
-
 </Project>
--- a/src/Microsoft.Repl.Tests/Microsoft.Repl.Tests.csproj
+++ b/src/Microsoft.Repl.Tests/Microsoft.Repl.Tests.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.0</TargetFramework>
@@ -10,6 +10,10 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.Repl\Microsoft.Repl.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="Parsing\" />
   </ItemGroup>
 
 </Project>
--- a/src/Microsoft.Repl/Parsing/CoreParseResultExtensions.cs
+++ b/src/Microsoft.Repl/Parsing/CoreParseResultExtensions.cs
@@ -1,0 +1,64 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Repl.Parsing
+{
+    public static class CoreParseResultExtensions
+    {
+        public static bool ContainsExactly(this ICoreParseResult parseResult, int length, StringComparison stringComparison, params string[] sections)
+        {
+            if (parseResult.Sections.Count != length)
+            {
+                return false;
+            }
+
+            return CompareSections(parseResult.Sections, sections, stringComparison);
+        }
+
+        public static bool ContainsExactly(this ICoreParseResult parseResult, int length, params string[] sections)
+        {
+            return ContainsExactly(parseResult, length, StringComparison.OrdinalIgnoreCase, sections);
+        }
+
+        public static bool ContainsExactly(this ICoreParseResult parseResult, params string[] sections)
+        {
+            return ContainsExactly(parseResult, sections.Length, sections);
+        }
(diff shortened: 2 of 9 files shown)


src/Microsoft.Repl.Tests/Microsoft.Repl.Tests.csproj
```diff
@@ -12,4 +12,8 @@
     <ProjectReference Include="..\Microsoft.Repl\Microsoft.Repl.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
jimmylewis: Not necessary?

src/Microsoft.Repl/Parsing/CoreParseResultExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        public static bool ContainsAtLeast(this ICoreParseResult parseResult, params string[] sections)
+        {
+            return ContainsAtLeast(parseResult, minimumLength: sections.Length, sections);
+        }
+
+        private static bool CompareSections(IReadOnlyList<string> parseSections, string[] sections, StringComparison stringComparison)
+        {
+            for (int index = 0; index < sections.Length; index++)
```
jimmylewis: A single indexer over two collections is usually a recipe for an IndexOutOfRangeException.  Check for `index < parseSections.Count` as well (or `Math.Min(parseSections.Count, sections.Length)`).

Actually, this raises a question: does this compare that all sections are equal, or just that the initial N sections are equal?  Because it could also be addressed by checking whether  `parseSections.Count == sections.Length` if you're looking for an exact match.

src/Microsoft.Repl.Tests/Microsoft.Repl.Tests.csproj
tlmii: Whoops. Yeah. Gone in 64653c2

src/Microsoft.Repl/Parsing/CoreParseResultExtensions.cs
tlmii: The fact that you're asking makes me think the methods aren't clear enough, so I might need to rethink. But to go ahead and explain: 

Probably easiest to look at the three-parameter overloads:
`ContainsExactly(this ICoreParseResult parseResult, int length, params string[] sections)` does two checks: it makes sure that parseResult has exactly `length` sections and that the first `sections.Length` of those sections matches the ones in the params arg. The common usage of this is the two-parameter overload `ContainsExactly(this ICoreParseResult parseResult, params string[] sections)` where `length` in the other overload is defaulted to `sections.Length`. In which case you're checking to make sure parseResult has exactly `sections.Length` sections and all of them match the sections in `params`.

Similarly, `ContainsAtLeast(this ICoreParseResult parseResult, int length, params string[] sections)` does two checks: it makes sure that parseResult has at least `length` sections and that the first `sections.Length` of those sections match the ones in the params arg. the common usage of this is again the two-parameter overload `ContainsAtLeast(this ICoreParseResult parseResult, params string[] sections)` where `length` in the other overload is defaulted to `sections.Length`. In which case you're checking to make sure parseResult has at least `sections.Length` sections and the first `sections.Length` of them match the sections in `params`.

The `CompareSections` function is just there to extract the duplicated logic between those two - the second check is exactly the same between the two cases, regardless of the first check. So I wrote the private function assuming that it is being called by the public functions and thus the length checks are already done there, no need to do them again. Except I didn't really do all the length checks that were necessary, so I just added those in aa1db28.

PR #84, opened 2019-07-15
Attempting to add testing on helix

tlmii:
This PR will probably never get committed. Just looking at moving the testing to use helix to see if it is easier/makes more sense than what we've currently got.


tlmii:
Closing this PR for now.

PR #98, opened 2019-07-18
Adding tests for Suggestions

tlmii:


src/Microsoft.HttpRepl.Tests/Suggestions/HeaderCompletionTests.cs
```diff
@@ -0,0 +1,196 @@
...
+            Assert.Contains("Upgrade-Insecure-Requests", result, StringComparer.OrdinalIgnoreCase);
+            Assert.Contains("User-Agent", result, StringComparer.OrdinalIgnoreCase);
+        }
+
+        [Fact]
+        public void GetCompletions_NoResults_ReturnsNull()
+        {
+            IEnumerable<string> result = HeaderCompletion.GetCompletions(null, "Not-A-Real-Header");
+
+            Assert.Null(result);
```
jimmylewis: spec question: would we rather return (and therefore have to be aware of) null, or would it be better to return an empty collection?

tlmii: A quick glance at the code makes it seem like returning an empty collection would be fine as well. The code in both places where this is used is written to handle both scenarios. I think we could simplify it to return an empty collection now that we have the tests to be able to assert that it doesn't return null. Let me take a little deeper look and I'll make a call.

tlmii: OK, I changed that up in 425e081.

src/Microsoft.HttpRepl/Suggestions/HeaderCompletion.cs
```diff
@@ -65,36 +65,24 @@ public class HeaderCompletion
...
-                    }
+                    IEnumerable<string> results = programState.GetApplicableContentTypes(method, path);
 
-                    return results.Where(x => !string.IsNullOrEmpty(x) && x.StartsWith(prefix, StringComparison.OrdinalIgnoreCase));
+                    return results?.Where(x => !string.IsNullOrEmpty(x) && x.StartsWith(prefix, StringComparison.OrdinalIgnoreCase));
                 default:
                     return null;
             }
-        }
+        }
```
jimmylewis: nit: clean up trailing whitespace

tlmii: Done in b2519c5.

PR #101, opened 2019-07-18
Clean up unit tests to not use SampleApi

bhsubra:
As a part of this pr:
1. Deleted HttpCommandTests and updated the tests to use CommandTestsBase instead
2.  Updated the unit tests to not use SampleApi. Instead they not use MockHttpHandler
3.  Updated tests modified in 1 and 2 to follow AAA pattern


src/Microsoft.HttpRepl.Fakes/MockHttpMessageHandler.cs
```diff
@@ -9,16 +11,46 @@ namespace Microsoft.HttpRepl.Fakes
 {
     public class MockHttpMessageHandler : HttpMessageHandler
     {
-        private HttpResponseMessage response;
+        private string _fileContents;
+        private string _filePath;
```
tlmii: `_filePath` is never used after it is set. Is that a mistake, or can it be removed?

src/Microsoft.HttpRepl.Fakes/MockHttpMessageHandler.cs
```diff
@@ -9,16 +11,46 @@ namespace Microsoft.HttpRepl.Fakes
 {
     public class MockHttpMessageHandler : HttpMessageHandler
     {
-        private HttpResponseMessage response;
+        private string _fileContents;
```
tlmii: Should all of these be `readonly`?

src/Microsoft.HttpRepl.Tests/Commands/CommandTestsBase.cs
```diff
@@ -1,12 +1,12 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Collections.Generic;
 using System.Net.Http;
 using Microsoft.HttpRepl.Fakes;
 using Microsoft.HttpRepl.FileSystem;
```
tlmii: This using appears to not be necessary anymore.

src/Microsoft.HttpRepl.Tests/Commands/PatchCommandTests.cs
```diff
@@ -1,94 +1,160 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Text;
+using System.Collections.Generic;
+using System.IO;
```
tlmii: I don't think this using is used anymore

src/Microsoft.HttpRepl.Fakes/MockHttpMessageHandler.cs
bhsubra: Yes, it not used. Removed it here:
https://github.com/aspnet/HttpRepl/commit/e438bd54448a497c75a978e6bd7cc95441349948

Thanks!

src/Microsoft.HttpRepl.Tests/Commands/PatchCommandTests.cs
bhsubra: Removed it here:
https://github.com/aspnet/HttpRepl/commit/41e90b3a9ce64cf9eafa0e461e37e7bbb4cf979c

src/Microsoft.HttpRepl.Tests/Commands/CommandTestsBase.cs
bhsubra: Its coming from IDictionary in ln 28

tlmii: I'm referring to `using Microsoft.HttpRepl.FileSystem;`

src/Microsoft.HttpRepl.Fakes/MockHttpMessageHandler.cs
bhsubra: Yes, updated here:
https://github.com/aspnet/HttpRepl/commit/6b03992aef4b9c8a1c2eebec091249e5b3e1069d

Thanks!

src/Microsoft.HttpRepl.Tests/Commands/CommandTestsBase.cs
bhsubra: Ah sorry. I removed it in one of the previous commits:
https://github.com/aspnet/HttpRepl/commit/f7d6cae5e54ca2564288d5df9cd085c907be4ac8

PR #103, opened 2019-07-19
Addressing issues in OpenApiV3EndpointMetadataReader

tlmii:
As I started doing additional integration tests yesterday, I realized that the V3 metadata reader was not handling most of the GET endpoints in our sample api server. For example, our sample api server includes an `HttpGet` route for `api/values`, but the current output of our swagger integration tests shows that it's not being picked up:

```
[BaseUrl]/api~ ls
.        []
..       []
Values   [post]
```

Digging a little further, it seems that the `ReadMetadata` method on `OpenApiV3EndpointMetadataReader` would only add a request method to the collection if it had a request body. Even before reading the spec, I realized that couldn't be right - `GET` apis aren't going to have a request body. And indeed, in our sample api swagger endpoint, the `GET` apis didn't have `requestBody` properties.

So that led me down the path of reading the OpenApi V3 spec, which you can find [here](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md) and [here](https://swagger.io/specification/). 

It turns out that for a request method (referred to in the spec as an operation), only the `responses` property is required and `requestBody` is optional:

`responses`:
> **REQUIRED**. The list of possible responses as they are returned from executing this operation.

`requestBody`:
> The request body applicable for this operation. The requestBody is only supported in HTTP methods where the HTTP 1.1 specification RFC7231 has explicitly defined semantics for request bodies. In other cases where the HTTP spec is vague, requestBody SHALL be ignored by consumers.

 The `content` property of the `requestBody` is required if the `requestBody` is present, but that doesn't make the `requestBody` required.

`content`:
> **REQUIRED**. The content of the request body. The key is a media type or media type range and the value describes it. For requests that match multiple keys, only the most specific key is applicable. e.g. text/plain overrides text/*

So, given all that, we have some choices to make in the code. For starters:

1. What do we do with the rest of the information in a particular `operation`/`requestMethod` if it has no `responses` property? We don't _need_ the `responses` property for what we're doing right now. But not having it would make it an invalid `operation`.
2. What do we do with the rest of the information in a particular `operation`/`requestMethod` if the `requestBody` property doesn't have a `content` property? We need the `content` property to populate the allowed content types, but we don't need it to be able to say the endpoint exists or that the `operation`/`requestMethod` exists.

In this PR, I went with the more restrictive answer to 1 (if there's no `responses` property, just skip that `operation` altogether, regardless of whatever else is there) and the more permissive answer to 2 (if there's no `content` property, leave the `operation` in place, but just don't have any allowed content types.

Happy to discuss a different set of answers if others disagree.

Along the way in this PR I also ended up changing some of the test data in the `OpenApiV3EndpointMetadataReaderTests` class because it was invalid. Mostly, this was a case of the `summary`, `operationId` and `responses` properties being inside `content` instead of inside the `operation` where they should be.

### DIFF
--- a/src/Microsoft.HttpRepl.Tests/OpenApi/OpenApiV3EndpointMetadataReaderTests.cs
+++ b/src/Microsoft.HttpRepl.Tests/OpenApi/OpenApiV3EndpointMetadataReaderTests.cs
@@ -112,6 +112,89 @@ namespace Microsoft.HttpRepl.Tests.OpenApi
         }
 
         [Fact]
+        public void ReadMetadata_WithContentAndOneContentType_ReturnsEndpointMetadataWithContentType()
+        {
+            string json = @"{
+  ""openapi"": ""3.0.0"",
+  ""paths"": {
+    ""/pets"": {
+      ""post"": {
+        ""summary"": ""Create a pet"",
+        ""operationId"": ""createPets"",
+        ""responses"": {
+          ""201"": {
+            ""description"": ""Null response""
+          }
+        },
+        ""requestBody"": {
+          ""description"": ""A Request Body"",
+          ""required"": false,
+          ""content"": {
+            ""application/json"": {
+            }
+          }
+        }
+      }
+    }
+  }
+}";
+            JObject jobject = JObject.Parse(json);
+            OpenApiV3EndpointMetadataReader openApiV3EndpointMetadataReader = new OpenApiV3EndpointMetadataReader();
+
+            List<EndpointMetadata> endpointMetadata = openApiV3EndpointMetadataReader.ReadMetadata(jobject).ToList();
+
+            Assert.Single(endpointMetadata);
+            Assert.Equal("/pets", endpointMetadata[0].Path);
+            Assert.Single(endpointMetadata[0].AvailableRequests);
+            Assert.True(endpointMetadata[0].AvailableRequests.ContainsKey("post"));
+            Assert.Single(endpointMetadata[0].AvailableRequests["post"]);
+            Assert.True(endpointMetadata[0].AvailableRequests["post"].ContainsKey("application/json"));
+        }
+
+        [Fact]
+        public void ReadMetadata_WithContentAndMultipleContentTypes_ReturnsEndpointMetadataWithContentTypes()
+        {
+            string json = @"{
+  ""openapi"": ""3.0.0"",
+  ""paths"": {
+    ""/pets"": {
+      ""post"": {
+        ""summary"": ""Create a pet"",
+        ""operationId"": ""createPets"",
+        ""responses"": {
+          ""201"": {
+            ""description"": ""Null response""
+          }
+        },
+        ""requestBody"": {
+          ""description"": ""A Request Body"",
+          ""required"": false,
+          ""content"": {
+            ""application/json"": {
+            },
+            ""text/plain"": {
+            }
+          }
+        }
+      }
+    }
+  }
+}";
+            JObject jobject = JObject.Parse(json);
+            OpenApiV3EndpointMetadataReader openApiV3EndpointMetadataReader = new OpenApiV3EndpointMetadataReader();
+
+            List<EndpointMetadata> endpointMetadata = openApiV3EndpointMetadataReader.ReadMetadata(jobject).ToList();
+
+            Assert.Single(endpointMetadata);
+            Assert.Equal("/pets", endpointMetadata[0].Path);
+            Assert.Single(endpointMetadata[0].AvailableRequests);
+            Assert.True(endpointMetadata[0].AvailableRequests.ContainsKey("post"));
+            Assert.Equal(2, endpointMetadata[0].AvailableRequests["post"].Count);
+            Assert.True(endpointMetadata[0].AvailableRequests["post"].ContainsKey("application/json"));
+            Assert.True(endpointMetadata[0].AvailableRequests["post"].ContainsKey("text/plain"));
+        }
+
+        [Fact]
         public void ReadMetadata_WithValidInput_ReturnsEndpointMetadata()
         {
             string json = @"{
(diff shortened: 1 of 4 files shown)


src/Microsoft.HttpRepl.Tests/OpenApi/OpenApiV3EndpointMetadataReaderTests.cs
```diff
@@ -77,21 +77,23 @@ public void ReadMetadata_WithNoRequestBody_ReturnsEndpointMetadataWithEmptyAvail
         }
 
         [Fact]
-        public void ReadMetadata_WithNoContent_ReturnsEndpointMetadataWithEmptyAvailableRequests()
+        public void ReadMetadata_WithNoContent_ReturnsEndpointMetadataWithRequestButNoContentTypes()
```
bhsubra: Is this a valid test case?
ReadMetadata_WithContent_ReturnsEndpointMetadataWithRequestAndContentTypes

tlmii: I added a couple tests for WithContent - when there's one and when there's more than one - in 9f1d12d.

src/Microsoft.HttpRepl/OpenApi/OpenApiV3EndpointMetadataReader.cs
```diff
@@ -36,6 +36,15 @@ public IEnumerable<EndpointMetadata> ReadMetadata(JObject document)
 
                         if (method.Value is JObject methodBody)
                         {
+                            if (!(methodBody["responses"] is JObject))
```
bhsubra: nit: Do you think it will be useful to add the links that you shared in the pr description here for future reference?

jimmylewis:
The decisions for the two points you call out seem reasonable, and I'm good with them.  Just trying to wrap my head around edge case scenarios though, I do question item 1 a bit:

>1. What do we do with the rest of the information in a particular operation/requestMethod if it has no responses property? We don't need the responses property for what we're doing right now. But not having it would make it an invalid operation.

>I went with the more restrictive answer to 1 (if there's no responses property, just skip that operation altogether, regardless of whatever else is there)

This is reasonable stance as long as we assume that:
a) the framework will encourage developers to include response details in their annotations, so that the API can be generated with the right data (i.e. they'll get a warning elsewhere), and 
b) this tool (HttpRepl) will be used to validate only endpoints which do adhere to the protocol.  To put it another way, should HttpRepl show a warning on an `operation` that does not adhere to the protocol, instead of silently skipping it?  Would a developer who is testing their API have a better experience trying to determine why it doesn't show up in the REPL vs. listing the `operation` with a warning?

Similarly on point 2, should we be showing a warning to the user if the endpoint is there but doesn't contain the required metadata?

In both these cases, the user can still invoke the API and see the response, since we don't *need* endpoint data to use the REPL.  So if we have any data at all indicating that the `operation` exists, we should show it - it's still a productivity boost over ignoring it.

tlmii:
> the framework will encourage developers to include response details in their annotations, so that the API can be generated with the right data

I guess it depends on what you mean by the framework. In the case of using Swashbuckle to generate the api spec, it always includes a `responses` property with an HTTP 200 response even if the developer didn't specify return types in other ways. You can see that in their code [here](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/07adbdbf833510bfb050b98da3a037c2f911722b/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/SwaggerGenerator.cs#L363). That was part of what I based my decision on - for our primary scenario (people using this to test ASP.NET Core Web APIs with the recommended swagger generation tool of Swashbuckle - they'll always have a `responses` property. 

>  this tool (HttpRepl) will be used to validate only endpoints which do adhere to the protocol.

This answer is probably similar. We of course can't guarantee it'll only be used to validate endpoints which adhere to the protocol. But the default scenario will adhere to it (at least to the extent that Swashbuckle continues to). 

I think we could easily build on top of this later by adding in warnings about things or being more explicit about where we drop endpoints or whatever. But I think it makes the most sense to start off being restrictive on point 1 and then add warnings (in both scenarios) rather than trying to accept it now and back away later (though I guess we could do that type of change with the next major version).

src/Microsoft.HttpRepl/OpenApi/OpenApiV3EndpointMetadataReader.cs
tlmii: Added it in https://github.com/aspnet/HttpRepl/pull/103/commits/f19bd21274b4886d7153e60444bc1f17349da915

jimmylewis:
I was being a little bit vague intentionally about "the Framework", because while this tool is written in .NET and aimed primarily at .NET developers, it could be used by anybody testing a REST API.  So, it could be .NET with Swashbuckle, or Node with swagger-node, or Rails with swagger-docs, or...

Considering purely from this tool's perspective, it seems like it should give a best effort to show any details it can, even incomplete or invalid APIs, so that the developer can be notified of what the issues may be and take steps to address them.  If there were hypothetical bug in a new version of Swagger that misses one of these fields, what should the user experience be?

tlmii:
I think providing as much info as we can glean plus warnings about things that are missing or wrong would likely be the right end goal. The question for right now is what about now, when we don't have the warnings or the proper setup in the code/data to provide them?

I think I've changed my mind a bit about being restrictive. If we allow it now, then the addition of warnings later just comes with the warnings and no additional endpoints just suddenly showing up. If we  disallow it now, and then allow it with a warning later, that'll change the output. 

As long as we are limiting it to where we don't enforce required sections, rather than trying to handle non-standard scenarios (like they use an array sometimes instead of an object), I'm comfortable being more permissive for this type of tool.

jimmylewis:
>As long as we are limiting it to where we don't enforce required sections, rather than trying to handle non-standard scenarios (like they use an array sometimes instead of an object), I'm comfortable being more permissive for this type of tool.

This sounds great.  Missing data is one thing, returning an unexpected format or data type is another.

PR #106, opened 2019-07-19
ListCommand and GetCommand Integration Tests

tlmii:
First stab at adding more integration tests. I converted the existing `SwaggerIntegrationTests` and `NonSwaggerIntegrationTests` over to be `ListCommandTests `and then added `GetCommandTests`.

I also removed the old routes for the old integration tests that we aren't using anymore. For whatever reason, those were causing the sample api server not to respond to what should have been valid api requests for routes on ValuesController.

If these tests look like a decent start, I'll add more tests for other commands in a subsequent PR.


src/Microsoft.HttpRepl.IntegrationTests/Commands/GetCommandTests.cs
```diff
@@ -0,0 +1,176 @@
...
+            output = NormalizeOutput(output, _serverConfig.BaseAddress);
+
+            string expected = NormalizeOutput(@"(Disconnected)~ set base [BaseUrl]
+Using swagger metadata from [BaseUrl]/swagger/v1/swagger.json
+
+[BaseUrl]/~ cd api/values
+/api/values    [get|post]
+
+[BaseUrl]/api/values~ get
+[32m[1mHTTP[22m[39m[32m[1m/[22m[39m[32m[1m1.1[22m[39m [33m[1m200[22m[39m [33m[1mOK[22m[39m
```
jimmylewis: I'm assuming the gibberish here are the console color codes.  If I have alternate colors configured, will it affect running tests on my machine?  Should we (or even can we) normalize these as well?

(For that matter, will we eventually have integration tests for configuring colors...)

src/Microsoft.HttpRepl.IntegrationTests/Commands/GetCommandTests.cs
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        [Fact]
+        public async Task WithoutParameter_ShowsColorizedOutput()
+        {
+            string scriptText = $@"set base {_serverConfig.BaseAddress}
+cd api/values
+get";
+
+            var console = new LoggingConsoleManagerDecorator(new NullConsoleManager());
```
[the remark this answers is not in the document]
tlmii: Done in https://github.com/aspnet/HttpRepl/pull/106/commits/5f8d3e0b867291dc3b87f6c7dabbbca09c96170e.

src/Microsoft.HttpRepl.IntegrationTests/Commands/GetCommandTests.cs
tlmii: Yes, your configuration would impact these tests as far as I can tell. Normalizing it might be tricky... the same color can be from different sources, so it's not really reversible. 

I _would_ like to have integration tests for configuring the colors, especially since in some cases there's no color codes in the output (if you've specified None). So the output looks different than the default both raw and rendered. 

I'll ponder that a little bit more - we need a solution since the colors are going to be in everything that has output from a server, plus errors and warnings.

jimmylewis: One option might be to have a set of preferences that we apply during tests to normalize the output (whether part of the test script, or an injected `IPreferences`).  That at least alleviates the variability based on local settings.

Aside from tests actually around color settings, I wonder if we'd lose anything by setting all colors to none (i.e. preventing getting these codes in the output).  It would mean that tests assert the right textual output, but no longer enforce the right colorization.  On a select few tests, we can re-enable colors to make sure we're doing the right thing.  Thoughts?

tlmii: I thought about your second paragraph a good bit last night and I just can't really decide. The colors are kind of an intrinsic part of the output - whether it is the json colorization of output or warning/error colorization of messages.

Maybe there's a middle ground where we keep the error/warning colors, since those will pollute the output a lot less but make all the others None? The error/warning colors are truly part of the message in my mind... but the json colorization (or xml colorization in the future) is just a readability helper. But that might be a bit of post-decision reduction on my part.

Issue #112, opened 2019-07-25
HttpClient with DefaultCredentials


MarcusKohnert:
As I'm working in an enterprise environment I would like to be able to turn on `UseDefaultCredentials=true` on `HttpClient`.

Most of our HTTP endpoints use Kerberos Authentication (aka WindowsAuthentication in IIS). Those endpoints require that the HttpClient sends proper user authentication in it's requests.

tlmii:
Closing this since the change was merged already.

PR #113, opened 2019-07-25
Enable preferences for HttpClient

MarcusKohnert:
note first: HttpRepl is awesome!

Draft to solve #112

I've created this early draft PR to get early feedback from you if that scenario is even possible and to enable guidance on how this feature might be implemented.


src/Microsoft.HttpRepl/Preferences/HttpClientPreferences.cs
```diff
@@ -0,0 +1,38 @@
...
+                            matchingProperties.Add(val);
+                        }
+                    }
+
+                    return _names = matchingProperties;
+                }
+            }
+        }
+
+        public static string UseDefaultCredentials { get; } = "httpClient.useDefaultCredentials";
```
tlmii: For now, its probably safe to just add this to `WellKnownPreference` rather than making the separate `HttpClientPreferences`. We may at some point split them out logically (colors, formatting, editor, swagger, etc) but until we know exactly how we're going to do that, I'd lean towards just keeping it in `WellKnownPreference`.

src/Microsoft.HttpRepl/Program.cs
```diff
@@ -120,5 +121,16 @@ private static void ComposeDependencies(IConsoleManager consoleManager, out Http
                 { WellKnownPreference.JsonColor, "Green" }
             };
         }
+
+        private static HttpClient GetHttpClientWithPreferences(IPreferences preferences)
+        {
+            if (preferences.TryGetValue(HttpClientPreferences.UseDefaultCredentials, out var value) &&
```
tlmii: I'm somewhat surprised that this is the first preference we have that is a boolean, but it seems that it is. We should probably go ahead and add a `bool GetBoolValue(string preference, bool defaultValue)` to `IPreferences` and its implementations and use that here.

src/Microsoft.HttpRepl.Fakes/NullPreferences.cs
```diff
@@ -22,7 +22,12 @@ public int GetIntValue(string preference, int defaultValue = default)
             return defaultValue;
         }
 
-        public string GetValue(string preference, string defaultValue = default)
```
tlmii: We just pushed a change to this yesterday, changing the `string defaultValue = null` to `string defaultValue = default`. Guessing the PRs just crossed paths. Can you change that back here?

MarcusKohnert: sorry for missing that

tlmii: No worries. Thanks for the quick reply!

PR #115, opened 2019-07-25
Refactored BaseHttpCommand and added couple of tests to verify format behavior

bhsubra:
As a part of this pr:
- Refactored BaseHttpCommand class to make it more readable
- Added couple of tests to verify format behavior based on content type
- I did not find more areas for testing as most of the areas are covered by Commands that implement the abstract methods. Let me know if you have any suggestions around adding more tests

### DIFF
--- a/src/Microsoft.HttpRepl.Fakes/MockHttpMessageHandler.cs
+++ b/src/Microsoft.HttpRepl.Fakes/MockHttpMessageHandler.cs
@@ -11,6 +11,7 @@ namespace Microsoft.HttpRepl.Fakes
 {
     public class MockHttpMessageHandler : HttpMessageHandler
     {
+        private readonly string _contentType;
         private readonly string _fileContents;
         private readonly string _header;
         private readonly bool _readFromFile;
@@ -19,8 +20,10 @@ namespace Microsoft.HttpRepl.Fakes
         public MockHttpMessageHandler(IDictionary<string, string> urlsWithResponse,
             string header,
             bool readFromFile,
-            string fileContents)
+            string fileContents,
+            string contentType)
         {
+            _contentType = contentType;
             _fileContents = fileContents;
             _header = header;
             _readFromFile = readFromFile;
@@ -45,6 +48,11 @@ namespace Microsoft.HttpRepl.Fakes
             else
             {
                 httpResponseMessage.Content = new MockHttpContent(responseContent);
+            }
+
+            if (!string.IsNullOrEmpty(_contentType) && httpResponseMessage.Content != null)
+            {
+                httpResponseMessage.Content.Headers.Add("Content-Type", _contentType);
             }
 
             return Task.FromResult(httpResponseMessage);
--- a/src/Microsoft.HttpRepl.Fakes/MockHttpMessageHandler.cs
+++ b/src/Microsoft.HttpRepl.Fakes/MockHttpMessageHandler.cs
@@ -36,6 +36,7 @@ namespace Microsoft.HttpRepl.Fakes
             _urlsWithResponse.TryGetValue(absoluteUri, out string responseContent);
 
             HttpResponseMessage httpResponseMessage = new HttpResponseMessage(HttpStatusCode.OK);
+            httpResponseMessage.RequestMessage = request;
 
             if (_readFromFile)
             {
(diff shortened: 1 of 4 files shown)


src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
```diff
@@ -118,175 +113,159 @@ protected override async Task ExecuteAsync(IShellState shellState, HttpState pro
...
-                }
-
-                consoleManager.WriteLine();
-                consoleManager.WriteLine($"Response from {hostString}...".SetColor(requestConfig.AddressColor));
-                consoleManager.WriteLine();
-                foreach (string responseLine in responseOutput)
-                {
-                    consoleManager.WriteLine(responseLine);
-                }
+                protocolInfo = await HandleEchoRequest(commandInput, consoleManager, programState, response, cancellationToken);
```
tlmii: Something doesn't seem right here. `protocolInfo` is set here, but then will be almost immediately overwritten on L269. It looks like it is used in `HandleEchoRequest`. Does it just not need to be returned?

bhsubra: Ah, actually we don't need to return it. It just needs to be written to console manager. Updated here:
https://github.com/aspnet/HttpRepl/commit/a41ab80846f7a8137b96f77c1e2a0b8cb2b83e34

src/Microsoft.HttpRepl.Tests/Commands/GetCommandTests.cs
```diff
@@ -97,5 +97,85 @@ public async Task ExecuteAsync_WithOnlyBaseAddress_VerifyOutput()
...
+  },
+  ""paths"": {
+    ""/api/Employees"": {
+    }
+  }
+}";
+            List<string> result = shellState.Output;
+
+            Assert.Equal(3, result.Count);
+            Assert.Contains("HTTP/1.1 200 OK", result);
```
jimmylewis: Should we be asserting about the order of these?  E.g. use `Assert.Equal("HTTP/1.1 200 OK", result[0]);`

src/Microsoft.HttpRepl.Fakes/MockHttpMessageHandler.cs
```diff
@@ -19,8 +20,10 @@ public class MockHttpMessageHandler : HttpMessageHandler
         public MockHttpMessageHandler(IDictionary<string, string> urlsWithResponse,
             string header,
             bool readFromFile,
-            string fileContents)
+            string fileContents,
+            string contentType)
```
jimmylewis: It's a little odd to set the contentType on the MessageHandler, as that will apply to all messages from this handler.  However, since this is limited, it's probably fine - it'll only become a limitation when/if we have a test that has multiple responses with separate content types.  (At which point we might want to focus on the scope of the test being too large... maybe).  We'll cross that bridge when we come to it.

src/Microsoft.HttpRepl.Fakes/MockHttpMessageHandler.cs
```diff
@@ -19,8 +20,10 @@ public class MockHttpMessageHandler : HttpMessageHandler
         public MockHttpMessageHandler(IDictionary<string, string> urlsWithResponse,
```
tlmii: This feels weird to me to have a constructor that takes three distinct possibilities of use cases - only one of `urlsWithResponse` OR `header` OR `readFromFile/fileContents` will be used for any given MockHttpMessageHandler instance. I'm guessing this was just written to follow along with the code that calls it, which has all of those options as default parameters. But this class itself doesn't have default parameters, and that three option construct just feels weird. Maybe it doesn't matter right now, since its "just" a mock and we may need to refactor later anyway (for reasons @jimmylewis mentioned in the comment a few minutes ago)? Just wanted to call it out to see what people thought.

jimmylewis: I agree with this, just didn't have a good way of phrasing it.  Seems like this fake component should really be a `Dictionary<string, HttpResponseMessage>`.  We've moved some of the response message details to the overall MessageHandler type to apply to all responses (but in a unit test, how many responses are you actually issuing anyway?  Shouldn't be too many); and I don't understand the value of the readFromFile/fileContents pairing (why not just provide the response content?).

I think it's worth re-evaluating this fake in a separate PR.

src/Microsoft.HttpRepl.Fakes/MockHttpMessageHandler.cs
bhsubra: Created an issue to clean this up:
https://github.com/aspnet/HttpRepl/issues/120

src/Microsoft.HttpRepl.Tests/Commands/GetCommandTests.cs
bhsubra: Sure, updated here:
https://github.com/aspnet/HttpRepl/commit/11d66469fca28c5150c36e837ba5f754d35f38f1

PR #116, opened 2019-07-25
Remove colors in integration tests

tlmii:
This PR follows on from the discussion on https://github.com/aspnet/HttpRepl/pull/106#discussion_r306454633.

It cleans up the integration test output substantially by allowing a `NullPreferences` object to be used.

Unfortunately, due to https://github.com/aspnet/HttpRepl/blob/6ffda0536534c4c9311b0a82eb29402af371817e/src/Microsoft.HttpRepl/HttpState.cs#L21 and https://github.com/aspnet/HttpRepl/blob/6ffda0536534c4c9311b0a82eb29402af371817e/src/Microsoft.HttpRepl/HttpState.cs#L23 the warning and error colors were not blanked out as part of the `NullPreferences` object. So I had to make a change there as well to always return `AllowedColors.None` rather than always returning the passed-in default. It doesn't appear that this change causes any other unintended consequences, but I wanted to call it out just in case.

### DIFF
--- a/src/Microsoft.HttpRepl.Fakes/NullPreferences.cs
+++ b/src/Microsoft.HttpRepl.Fakes/NullPreferences.cs
@@ -14,7 +14,7 @@ namespace Microsoft.HttpRepl.Fakes
 
         public AllowedColors GetColorValue(string preference, AllowedColors defaultValue = AllowedColors.None)
         {
-            return defaultValue;
+            return AllowedColors.None;
         }
 
         public int GetIntValue(string preference, int defaultValue = 0)
--- a/src/Microsoft.HttpRepl.Fakes/NullPreferences.cs
+++ b/src/Microsoft.HttpRepl.Fakes/NullPreferences.cs
@@ -12,17 +12,17 @@ namespace Microsoft.HttpRepl.Fakes
         public IReadOnlyDictionary<string, string> CurrentPreferences => null;
         public IReadOnlyDictionary<string, string> DefaultPreferences => null;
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.HttpRepl.Fakes/NullPreferences.cs
```diff
@@ -14,7 +14,7 @@ public class NullPreferences : IPreferences
 
         public AllowedColors GetColorValue(string preference, AllowedColors defaultValue = AllowedColors.None)
         {
-            return defaultValue;
+            return AllowedColors.None;
```
jimmylewis: If we change AllowedColors.None to be 0x0, this could remain as `default`.  It would also make us comply with [CA1008](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1008-enums-should-have-zero-value?view=vs-2019):

>If an enumeration that has the FlagsAttribute applied defines a zero-valued member, its name should be 'None' to indicate that no values have been set in the enumeration. Using a zero-valued member for any other purpose is contrary to the use of the FlagsAttribute in that the AND and OR bitwise operators are useless with the member. ...

tlmii: I made the changes we discussed offline with making `AllowedColors` more closely align with the spec. But note that it was returning `defaultValue` (the parameter) before, not `default` (as it is now). That's what allowed the Warning/Error colors to sneak in.

Changing `GetIntValue` and `GetValue` to return `default` along with `GetColorValue` really breaks things so I left those alone.

PR #117, opened 2019-07-25
Convert Microsoft.Repl to a .NET Standard library and add pack

tlmii:
In support of recent interest in the underlying Microsoft.Repl library, this PR converts it to a .NET Standard 1.3 library (as opposed to .net core 2.1 and 3.0 libraries) to expand its reach, and set it to generate a nuget package.

I picked 1.3 based on the chart [here](https://docs.microsoft.com/en-us/dotnet/standard/net-standard#net-implementation-support) and the fact that .NET Framework 4.6 is the lowest version in mainstream support (I think) listed in that chart. I'm open to suggestions on a different target, though.

The two new package refs were just because of things that weren't natively supported in netstandard1.3, not because of any changes in the library.

I left out the `<PackageProjectUrl>` in `Microsoft.Repl.csproj` since we don't have documentation or a website specifically for the Repl and it'll just default to the github url if we leave it out (i think). Let me know if that's not correct.

### DIFF
--- a/src/Microsoft.Repl/Input/KeyHandlers.cs
+++ b/src/Microsoft.Repl/Input/KeyHandlers.cs
@@ -68,50 +68,17 @@ namespace Microsoft.Repl.Input
             inputManager.RegisterKeyHandler(ConsoleKey.F23, Unhandled);
             inputManager.RegisterKeyHandler(ConsoleKey.F24, Unhandled);
 
-            inputManager.RegisterKeyHandler(ConsoleKey.Applications, Unhandled);
-            inputManager.RegisterKeyHandler(ConsoleKey.Attention, Unhandled);
-            inputManager.RegisterKeyHandler(ConsoleKey.BrowserBack, Unhandled);
-            inputManager.RegisterKeyHandler(ConsoleKey.BrowserFavorites, Unhandled);
-            inputManager.RegisterKeyHandler(ConsoleKey.BrowserForward, Unhandled);
-            inputManager.RegisterKeyHandler(ConsoleKey.BrowserHome, Unhandled);
-            inputManager.RegisterKeyHandler(ConsoleKey.BrowserRefresh, Unhandled);
-            inputManager.RegisterKeyHandler(ConsoleKey.BrowserSearch, Unhandled);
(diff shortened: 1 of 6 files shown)


src/Microsoft.Repl/Input/KeyHandlers.cs
```diff
@@ -68,50 +68,17 @@ public static void RegisterDefaultKeyHandlers(IInputManager inputManager)
             inputManager.RegisterKeyHandler(ConsoleKey.F23, Unhandled);
             inputManager.RegisterKeyHandler(ConsoleKey.F24, Unhandled);
 
-            inputManager.RegisterKeyHandler(ConsoleKey.Applications, Unhandled);
```
jimmylewis: What happens if any of these keys get hit?  Does the REPL try to print a character for them?

I'm wondering if in addition to targeting netstandard1.3, we should also target netstandard2.0 and put these removed lines back inside of an `#if NETSTANDARD2_0`.

jimmylewis: (Or maybe `#if !NETSTANDARD1_3` would be better for future proofing)

tlmii: Yeah, you're right, it attempts to print out `\0`, so that's not ideal. 
Theoretically, this is a problem as they add new keys even beyond 2.0, right? I wonder if we need some other sort of catch all such as if the KeyChar isn't printable, and it's not a registered key handler, then just ignore it. Though defining printable is probably fraught. `IsLetterOrDigit || IsPunctuation || IsSymbol || Environment.NewLine || ' '` probably gets us close... but I'm not sure we can do that reliably.
Thoughts?

jimmylewis: Does `!char.IsControl(...)` handle these correctly?

tlmii: As discussed offline, `!char.IsControl( ... )` is not sufficient.

I added https://github.com/aspnet/HttpRepl/issues/119 to track the work we'd need to do to handle this. I think it is separate from this PR, this PR just makes it more obvious and important.

PR #127, opened 2019-07-29
Added integration tests for - EchoCommand, SetBaseCommand, SetHeaderCommand

bhsubra:
Added integration tests for - EchoCommand, SetBaseCommand, SetHeaderCommand

### DIFF
--- a/src/Microsoft.HttpRepl.IntegrationTests/Commands/EchoCommandTests.cs
+++ b/src/Microsoft.HttpRepl.IntegrationTests/Commands/EchoCommandTests.cs
@@ -1,0 +1,54 @@
+using System.Threading.Tasks;
+using Microsoft.HttpRepl.IntegrationTests.SampleApi;
+using Xunit;
+
+namespace Microsoft.HttpRepl.IntegrationTests.Commands
+{
+    public class EchoCommandTests : BaseIntegrationTest, IClassFixture<HttpCommandsFixture<SampleApiServerConfig>>
+    {
+        private readonly SampleApiServerConfig _serverConfig;
+
+        public EchoCommandTests(HttpCommandsFixture<SampleApiServerConfig> fixture)
+        {
+            _serverConfig = fixture.Config;
+        }
+
+        [Fact]
+        public async Task WithEchoOn_ShowsCorrectOutput()
+        {
+            string scriptText = $@"set base {_serverConfig.BaseAddress}
+echo on";
+
+            string output = await RunTestScript(scriptText, _serverConfig.BaseAddress);
+
+            string expected = NormalizeOutput(@"(Disconnected)~ set base [BaseUrl]
+Using swagger metadata from [BaseUrl]/swagger/v1/swagger.json
+
+[BaseUrl]/~ echo on
+Request echoing is on
+
+[BaseUrl]/~", null);
+
+            Assert.Equal(expected, output);
+        }
+
+        [Fact]
+        public async Task WithEchoOff_ShowsCorrectOutput()
+        {
+            string scriptText = $@"set base {_serverConfig.BaseAddress}
+echo off";
+
+            string output = await RunTestScript(scriptText, _serverConfig.BaseAddress);
+
+            string expected = NormalizeOutput(@"(Disconnected)~ set base [BaseUrl]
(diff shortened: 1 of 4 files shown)


src/Microsoft.HttpRepl.IntegrationTests/Commands/EchoCommandTests.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        private readonly SampleApiServerConfig _serverConfig;
+
+        public EchoCommandTests(HttpCommandsFixture<SampleApiServerConfig> fixture)
+        {
+            _serverConfig = fixture.Config;
+        }
+
+        [Fact]
+        public async Task WithEchoOn_ShowsCorrectOutput()
```
jimmylewis: Should these also do something to verify the echo, testing that the setting is working, not just that the message is written when being toggled?

src/Microsoft.HttpRepl.IntegrationTests/Commands/SetBaseCommandTests.cs
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using Microsoft.HttpRepl.IntegrationTests.SampleApi;
+using Xunit;
+
+namespace Microsoft.HttpRepl.IntegrationTests.Commands
+{
+    public class SetBaseCommandTests : BaseIntegrationTest, IClassFixture<DualHttpCommandsFixture<SampleApiServerConfig>>
```
jimmylewis: Add a `WithSwaggerAndUnknownUri`?  I think we show a warning in that case.

src/Microsoft.HttpRepl.IntegrationTests/Commands/SetBaseCommandTests.cs
```diff
@@ -0,0 +1,50 @@
...
+            string expected = NormalizeOutput(@"(Disconnected)~ set base [BaseUrl]
+Using swagger metadata from [BaseUrl]/swagger/v1/swagger.json
+
+[BaseUrl]/~", null);
+
+            Assert.Equal(expected, output);
+        }
+
+        [Fact]
+        public async Task WithoutSwaggerAndValidUri_ShowsCorrectOutput()
```
jimmylewis: Minor nit: I think the name `WithoutSwaggerAndValidUri_ShowsCorrectOutput` is a little of a misnomer, because without swagger data there isn't a way to determine valid/invalid URIs.  Everything is unknown, until you actually issue a request.

src/Microsoft.HttpRepl.IntegrationTests/Commands/EchoCommandTests.cs
jimmylewis: I guess it's covered elsewhere, in the SetHeaderCommandTests below.

bhsubra: Closing this. Let me know if we need anything here. Thanks!

src/Microsoft.HttpRepl.IntegrationTests/Commands/SetBaseCommandTests.cs
bhsubra: Updated the test names to WithSwaggerConfig_ShowsCorrectOutput and WithNonSwaggerConfig_ShowsCorrectOutput. Let me know if that sounds better.

src/Microsoft.HttpRepl.IntegrationTests/Commands/SetBaseCommandTests.cs
bhsubra: Looks like we write to console manager error instead of writing to console manager directly:
https://github.com/aspnet/HttpRepl/blob/507b52c18ddabdfa0491cad2cd4e664c4146f82c/src/Microsoft.HttpRepl/Commands/SetBaseCommand.cs#L41

So added unit test for the scenario:
https://github.com/aspnet/HttpRepl/commit/736dd109181feaa37eadbb9959d5cf30aa38dcb6

PR #139, opened 2019-08-05
Fix the exception "NullReferenceException" throw by set swagger command.

littlesmilelove:
- This issue is cause by `Schema.Properties` is null,  when this exception occur, it will set the `Structure` to null.


tlmii:
Hi @littlesmilelove - thanks for the PR! Would you mind adding a test or two for this scenario? You can probably follow the pattern in `SetSwaggerCommandTests.ExecuteAsync_WithoutChildDirectories_SetsHttpStateSwaggerStructure` and `SetSwaggerCommandTests.ExecuteAsync_WithChildDirectories_SetsHttpStateSwaggerStructureWithChildDirectorStuctureInfo`.

Issue #135, opened 2019-08-03
Crash on "ls <Tab>"


TheXDS:
When typing "ls " followed by a tab keystroke crashes the application.

Console dump:
```
Unhandled exception. System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
   at System.Collections.Generic.List`1.get_Item(Int32 index)
   at Microsoft.Repl.Commanding.CommandWithStructuredInputBase`2.Suggest(IShellState shellState, TProgramState programState, TParseResult parseResult) in /_/src/Microsoft.Repl/Commanding/CommandWithStructuredInputBase.cs:line 68
   at Microsoft.Repl.Commanding.DefaultCommandDispatcher`2.CollectSuggestions(IShellState shellState) in /_/src/Microsoft.Repl/Commanding/DefaultCommandDispatcher.cs:line 92
   at Microsoft.Repl.Suggestions.SuggestionManager.NextSuggestion(IShellState shellState) in /_/src/Microsoft.Repl/Suggestions/SuggestionManager.cs:line 39
   at Microsoft.Repl.Input.KeyHandlers.Tab(ConsoleKeyInfo keyInfo, IShellState state, CancellationToken cancellationToken) in /_/src/Microsoft.Repl/Input/KeyHandlers.cs:line 233
   at Microsoft.Repl.Input.InputManager.StartAsync(IShellState state, CancellationToken cancellationToken) in /_/src/Microsoft.Repl/Input/InputManager.cs:line 214
   at Microsoft.HttpRepl.Program.Start(String[] args, IConsoleManager consoleManager) in /_/src/Microsoft.HttpRepl/Program.cs:line 73
   at Microsoft.HttpRepl.Program.Main(String[] args) in /_/src/Microsoft.HttpRepl/Program.cs:line 24
   at Microsoft.HttpRepl.Program.<Main>(String[] args)
```
It happens when attempting to get suggestion on empty parameter input.

PR #140, opened 2019-08-05
Check for being in bounds of commandName before accessing item

Tadimsky:
If you type in `dir` and then hit tab, HttpRepl will crash.
It was happening as it was accessing an invalid index (1) in the the `commandName` list when the count was 1.
This adds a quick check to see if the `SelectedSection` is in bounds of the list before accessing the element.

### DIFF
--- a/src/Microsoft.Repl/Commanding/CommandWithStructuredInputBase.cs
+++ b/src/Microsoft.Repl/Commanding/CommandWithStructuredInputBase.cs
@@ -17,7 +17,7 @@ namespace Microsoft.Repl.Commanding
 
         public string GetHelpDetails(IShellState shellState, TProgramState programState, TParseResult parseResult)
         {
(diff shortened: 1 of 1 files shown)


Tadimsky:
Oh, it looks like this may be the issue reported in #135

tlmii:
@Tadimsky Thanks for the PR! Would you mind adding a test or two for this scenario as well?

Tadimsky:
@tlmii Sure, I'll give it a try! :)

tlmii:
After doing a little more digging into this issue, I've discovered that it was a little more complex than just the out of bounds issue at that particular line. It shouldn't have even been in that block of code to begin with. I've created #163 to address the issue so I'm going to close this PR.

Tadimsky:
Awesome, thanks @tlmii !

PR #143, opened 2019-08-05
README.md - Update Installation Command

DanRigby:
- Use quotes for the version parameter.
  - Without the quotes the install command will fail when running in ZSH with the following Error: `zsh: no matches found: 3.0.0*`

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ The HTTP Read-Eval-Print Loop (REPL) is:
 To install the HTTP REPL, run the following command:
 
 ```
-dotnet tool install -g Microsoft.dotnet-httprepl --version 3.0.0-*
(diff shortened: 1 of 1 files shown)


tlmii:
cc @scottaddie since we'll probably want this change on the docs too. This will be less relevant once released (since you won't have to specify the version number) but for now, might as well not make it more difficult on anyone testing.

DanRigby:
@tlmii Just realized I put the opening quote in the wrong place in this PR. Going to open a new PR to fix it.

DanRigby:
Opened #144 to fix it. The command should be:
`dotnet tool install -g Microsoft.dotnet-httprepl --version "3.0.0-*"`

Issue #148, opened 2019-08-06
HttpRepl doesn't respect servers:url as a base automatically


shanselman:
I've got a simple node app with a swagger.json like this.
[swagger.txt](https://github.com/aspnet/HttpRepl/files/3474294/swagger.txt)

`{
  "openapi": "3.0.0",
  "servers": [
    {
      "url": "/api/v1"
    }
  ]...`

And while it is correct and valid swagger, it's not automatically usable with httpRepl without manual and non-intuitive base url setting.

`set base https://MYSITE/api/v1
set swagger https://MYSITE/swagger.json`

The base setting should be implied by the single `servers:url` and this should work without manual set base or manual set swagger.

Second ask - Would be nice to "Set swagger LOCALFILE" as well. Or even "`dotnet watch httprep --swagger LOCALFILE`" for iterative development. 

cc @tlmii

tlmii:
The OpenAPI 3.0 spec info for the servers array can be found [here](https://swagger.io/docs/specification/api-host-and-base-path/) (examples in YAML only for whatever reason). But in general, it just looks like:

```json
{
  "openapi": "3.0.0",
  "servers": [
    {
      "url": "[relative or absolute url to the base address]",
      "description": "[optional markdown formatted description of the server's use case]"
    },
    {
      "url": "...",
      "description": "..."
    }
  ]
}
```

The 2.0 swagger spec had a more simplistic way of dealing with this that only supported a single base address. But it still allowed it to be specified. You can see some detail on that [here](https://swagger.io/docs/specification/2-0/api-host-and-base-path/). But again, generally, it looks like this:

```json
{
  "swagger": "2.0",
  "host": "example.com",
  "basePath": "api/v1",
  "schemes": [
    "https",
    "http"
  ]
}
```

The current implementation ignores both of these. Not only does this force the user to do redundant work in many cases (specifying the base address and swagger location when the latter already specifies the former), but in cases like Scott's it can be pretty opaque as to why it doesn't work. 

The way the tool is structured right now, it seems to be geared heavily towards the base address and extrapolating from that. But since the swagger file can specify a base address, it makes me wonder if that isn't backwards.

For instance, right now, the only (non-help) command line parameter for the tool is to pass it the base address:

`dotnet httprepl https://example.com/`

That essentially just calls `set base https://example.com/` immediately after the tool starts, which in turn, if the swagger file can be located, might also set the swagger file.

[...]

PR #161, opened 2019-08-12
Rework EndpointMetadataReader to support more of the swagger definition

tlmii:
This PR does some setup work to prepare for the scenarios mentioned in #148. These changes are all under the hood:

1. Previously, we just stored the directory structure (and method/content type info) from the swagger definition. Now, the new `ApiDefinition` class is a container for that information, the server information, and anything else we might want to track in the future.
2. Because of 1, I renamed a lot of things from \*EndpointMetadata\* to \*ApiDefinition\* to make the code more clear what it was doing now. Unfortunately, this has the effect of looking like I deleted/added a lot of files. 
3. From v2 specs, we construct a server url for each string listed in the `schemes` property. So if they have `"schemes": [ "http", "https" ]`, we're going to have two urls, one for each of those.

### DIFF
--- a/src/Microsoft.HttpRepl.Tests/Commands/ChangeDirectoryCommandTests.cs
+++ b/src/Microsoft.HttpRepl.Tests/Commands/ChangeDirectoryCommandTests.cs
@@ -23,7 +23,9 @@ namespace Microsoft.HttpRepl.Tests.Commands
 
             Setup(commandText: "cd NotAnEndpoint", out MockedShellState mockedShellState, out HttpState httpState, out ICoreParseResult parseResult);
 
-            httpState.Structure = new DirectoryStructure(null);
+            ApiDefinition apiDefinition = new ApiDefinition();
+            apiDefinition.DirectoryStructure = new DirectoryStructure(null);
+            httpState.ApiDefinition = apiDefinition;
 
             string expectedFirstLine = string.Format(Resources.Strings.ChangeDirectoryCommand_Warning_UnknownEndpoint, "/NotAnEndpoint").SetColor(httpState.WarningColor);
             string expectedSecondLine = "/NotAnEndpoint    []";
@@ -47,7 +49,9 @@ namespace Microsoft.HttpRepl.Tests.Commands
             RequestInfo childRequestInfo = new RequestInfo();
             childRequestInfo.AddMethod("GET");
             childDirectory.RequestInfo = childRequestInfo;
-            httpState.Structure = directoryStructure;
+            ApiDefinition apiDefinition = new ApiDefinition();
+            apiDefinition.DirectoryStructure = directoryStructure;
+            httpState.ApiDefinition = apiDefinition;
 
             string expectedOutput = "/AnEndpoint    [GET]";
 
@@ -67,7 +71,9 @@ namespace Microsoft.HttpRepl.Tests.Commands
             DirectoryStructure directoryStructure = new DirectoryStructure(null);
             DirectoryStructure childDirectory = directoryStructure.DeclareDirectory("AnEndpoint");
             DirectoryStructure grandchildDirectory = childDirectory.DeclareDirectory("AnotherEndpoint");
-            httpState.Structure = directoryStructure;
+            ApiDefinition apiDefinition = new ApiDefinition();
+            apiDefinition.DirectoryStructure = directoryStructure;
+            httpState.ApiDefinition = apiDefinition;
 
             string expectedOutput = "/AnEndpoint    []";
 
--- a/src/Microsoft.HttpRepl.Tests/Commands/ChangeDirectoryCommandTests.cs
+++ b/src/Microsoft.HttpRepl.Tests/Commands/ChangeDirectoryCommandTests.cs
@@ -24,8 +24,10 @@ namespace Microsoft.HttpRepl.Tests.Commands
 
             Setup(commandText: "cd NotAnEndpoint", out MockedShellState mockedShellState, out HttpState httpState, out ICoreParseResult parseResult);
 
-            ApiDefinition apiDefinition = new ApiDefinition();
-            apiDefinition.DirectoryStructure = new DirectoryStructure(null);
+            ApiDefinition apiDefinition = new ApiDefinition()
+            {
+                DirectoryStructure = new DirectoryStructure(null)
+            };
             httpState.ApiDefinition = apiDefinition;
 
             string expectedFirstLine = string.Format(Resources.Strings.ChangeDirectoryCommand_Warning_UnknownEndpoint, "/NotAnEndpoint").SetColor(httpState.WarningColor);
@@ -50,8 +52,10 @@ namespace Microsoft.HttpRepl.Tests.Commands
             RequestInfo childRequestInfo = new RequestInfo();
             childRequestInfo.AddMethod("GET");
             childDirectory.RequestInfo = childRequestInfo;
-            ApiDefinition apiDefinition = new ApiDefinition();
-            apiDefinition.DirectoryStructure = directoryStructure;
+            ApiDefinition apiDefinition = new ApiDefinition()
+            {
+                DirectoryStructure = directoryStructure
+            };
             httpState.ApiDefinition = apiDefinition;
 
             string expectedOutput = "/AnEndpoint    [GET]";
@@ -72,8 +76,10 @@ namespace Microsoft.HttpRepl.Tests.Commands
             DirectoryStructure directoryStructure = new DirectoryStructure(null);
             DirectoryStructure childDirectory = directoryStructure.DeclareDirectory("AnEndpoint");
             DirectoryStructure grandchildDirectory = childDirectory.DeclareDirectory("AnotherEndpoint");
-            ApiDefinition apiDefinition = new ApiDefinition();
-            apiDefinition.DirectoryStructure = directoryStructure;
+            ApiDefinition apiDefinition = new ApiDefinition()
+            {
+                DirectoryStructure = directoryStructure
+            };
             httpState.ApiDefinition = apiDefinition;
 
             string expectedOutput = "/AnEndpoint    []";
--- a/src/Microsoft.HttpRepl.Tests/Commands/ChangeDirectoryCommandTests.cs
+++ b/src/Microsoft.HttpRepl.Tests/Commands/ChangeDirectoryCommandTests.cs
@@ -6,6 +6,7 @@ using System.Threading;
 using Microsoft.HttpRepl.Commands;
 using Microsoft.HttpRepl.Fakes;
 using Microsoft.HttpRepl.FileSystem;
+using Microsoft.HttpRepl.OpenApi;
 using Microsoft.HttpRepl.Preferences;
 using Microsoft.HttpRepl.UserProfile;
 using Microsoft.Repl.ConsoleHandling;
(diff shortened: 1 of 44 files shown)


metoly:
Should we provide a way to choose (without having to type it out) from the list of server urls if multiple are present in the spec?
>> We can provide commands "list base/swagger" to help list the existing swagger files and server URLS while highlighting the selected ones with a "*".

Discussion with @jodavis earlier today brought up the topic of whether or not it is correct to overwrite what the user explicitly set in a set base call (for instance, if they do a set base with swagger.autoDetect being true, and we detect a swagger file with a different base address).
>> Explicit user commands should supersede settings and auto-detection.

Further to 2, we discussed whether it would be better to have a command, which for now I'll call set root, that if you use that, it does all the autodetection magic to set swagger and set base. But if you call set base, what you specify will never be changed unless you explicitly call set root.
>> Not sure about this one. If we give the user the ability to list, set and have an autodetect setting, "set root" would be redundant and might cause confusion down the road.

tlmii:
@metoly 
> Explicit user commands should supersede settings and auto-detection.

It sounds like this would mean that if the user ever called `set base`, then `set swagger` would never auto `set base`. 

That seems reasonable in most cases except that it forces the user to know the swagger url in order to get an automatically set base address.

One of the things that @jodavis and I discussed yesterday was how this ties into VS/VS Code launching the tool and the project system having to determine where things are to pass to the tool.

As it stands right now, VS/VS Code would just need to pass the root url pretending it is the base url (for instance, https://localhost:44301/) and the tool can autodetect the swagger file (at https://localhost:44301/swagger.json or wherever) and then autodetect the true correct base url (for instance https://localhost:44301/api/v1/).

But if we don't allow `set swagger` to `set base` automatically in that scenario, then either 
1. VS/VS Code would have to know to pass in https://localhost:44301/api/v1 and we would need new swagger detection logic to check the root for the swagger file, not just subpaths, or
2. We would need to implement a different set of arguments for the executable that allow the swagger url to be passed in, and VS/VS Code would have to know to pass in https://localhost:44301/swagger.json.

Either of those things other than the root are more complex for VS/VS Code to have to figure out to send us.

I guess a third option is to change the executable argument to be this "root" concept, not the base, and if it comes in via that, we can do the autodetect, but otherwise, `set base` never autodetects. That is closer to the `set root` idea above, but without having an explicit command the user would need to use.

metoly:
> @metoly
> 
> Explicit user commands should supersede settings and auto-detection.
> 
> It sounds like this would mean that if the user ever called set base, then set swagger would never auto set base.
> That seems reasonable in most cases except that it forces the user to know the swagger url in order to get an automatically set base address.

@tlmii: Today, I set base to http://localhost:44301 and swagger to https://localhost:44301/swagger.json, then everything works. Does this mean that when I set swagger, the base changes?

> One of the things that @jodavis and I discussed yesterday was how this ties into VS/VS Code launching the tool and the project system having to determine where things are to pass to the tool.
> As it stands right now, VS/VS Code would just need to pass the root url pretending it is the base url (for instance, https://localhost:44301/) and the tool can autodetect the swagger file (at https://localhost:44301/swagger.json or wherever) and then autodetect the true correct base url (for instance https://localhost:44301/api/v1/).
> But if we don't allow set swagger to set base automatically in that scenario, then either
> 
> VS/VS Code would have to know to pass in https://localhost:44301/api/v1 and we would need new swagger detection logic to check the root for the swagger file, not just subpaths, or
> We would need to implement a different set of arguments for the executable that allow the swagger url to be passed in, and VS/VS Code would have to know to pass in https://localhost:44301/swagger.json.
> 
> Either of those things other than the root are more complex for VS/VS Code to have to figure out to send us.
> I guess a third option is to change the executable argument to be this "root" concept, not the base, and if it comes in via that, we can do the autodetect, but otherwise, set base never autodetects. That is closer to the set root idea above, but without having an explicit command the user would need to use.

@tlmii: We can pass multiple parameters to the command, the base and Swagger URI.

tlmii:
@metoly: With this PR, if you set base and then swagger like you say, _and_ the swagger file points to a different base url for some reason, then the base would change (because your call to `set swagger` came second, and the swagger file you specified included a base address in addition to the directory structure). If the swagger file pointed to the same base address, or didn't include a base address, nothing would change. 

Personally, it seems like if you're specifying a swagger file, you are telling the tool to set itself up based on the contents of the swagger file. That would include the base address, the directory structure, the supported request methods and content types, etc. If you then want to override the base address that came with it (or any of the other pieces, really), you're free to do so.

metoly:
> @metoly: With this PR, if you set base and then swagger like you say, and the swagger file points to a different base url for some reason, then the base would change (because your call to set swagger came second, and the swagger file you specified included a base address in addition to the directory structure). If the swagger file pointed to the same base address, or didn't include a base address, nothing would change.

@tlmii: The problem is that set base is the command that establishes the connection. It might make more set sense to have connect, set base and set swagger then in this case.

> Personally, it seems like if you're specifying a swagger file, you are telling the tool to set itself up based on the contents of the swagger file. That would include the base address, the directory structure, the supported request methods and content types, etc. That doesn't seem like auto-detection (the way set base calls set swagger does) - it is an explicit user action. If you then want to override the base address that came with it (or any of the other pieces, really), you're free to do so.

I still believe that it would be confusing to have set swagger change an explicit user command. The explicit connect command might help, but there needs to be a clear explicit command to change several settings.

tlmii:
@metoly @jodavis Per our discussions this morning, I've streamlined this PR a bit to not have the user facing changes it originally had, but to still include the architectural changes that'll be helpful to the `connect` command  work we discussed.

src/Microsoft.HttpRepl.Tests/Commands/ChangeDirectoryCommandTests.cs
```diff
@@ -23,7 +24,9 @@ public void ExecuteAsync_UnknownEndpoint_DisplaysWarning()
 
             Setup(commandText: "cd NotAnEndpoint", out MockedShellState mockedShellState, out HttpState httpState, out ICoreParseResult parseResult);
 
-            httpState.Structure = new DirectoryStructure(null);
+            ApiDefinition apiDefinition = new ApiDefinition();
+            apiDefinition.DirectoryStructure = new DirectoryStructure(null);
```
jimmylewis: nit: use this pattern to initialize the object:
```
ApiDefinition apiDefinition = new ApiDefinition();
{
    DirectoryStructure = new DirectoryStructure(null),
};
```

src/Microsoft.HttpRepl.Tests/Commands/SetSwaggerCommandTests.cs
```diff
@@ -232,6 +233,40 @@ public async Task ExecuteAsync_WhenThirdSectionIsNotAValidUri_WritesToConsoleMan
...
+                urlsWithResponse: new Dictionary<string, string> { { "http://localhost/swagger.json", response } },
+                out MockedShellState shellState,
+                out HttpState httpState,
+                out ICoreParseResult parseResult,
+                out _,
+                out _);
+
+            SetSwaggerCommand setSwaggerCommand = new SetSwaggerCommand();
+
+            Assert.Equal("http://localhost/", httpState.BaseAddress.ToString(), StringComparer.Ordinal);
```
jimmylewis: is this worth asserting over?  Having a pre-assert kind of throws off AAA readability.

src/Microsoft.HttpRepl.Tests/OpenApi/ApiDefinitionReaderTests.cs
```diff
@@ -0,0 +1,49 @@
...
+            string json = @"{
+  ""info"": {
+    ""version"": ""v1"",
+    ""title"": ""My API""
+  }
+}";
+            JObject jobject = JObject.Parse(json);
+            ApiDefinitionReader apiDefinitionReader = new ApiDefinitionReader();
+
+            Assert.Null(apiDefinitionReader.Read(jobject, null));
```
jimmylewis: nit: for clearer AAA form, pull out the Act separate from the Assert:

```
var result = apiDefinitionReader.Read(jobject, null);

Assert.Null(result);
```

src/Microsoft.HttpRepl.Tests/OpenApi/SwaggerV1ApiDefinitionReaderTests.cs
```diff
@@ -0,0 +1,358 @@
...
+      ]
+    }
+  ]
+}";
+            JObject jobject = JObject.Parse(json);
+            SwaggerV1ApiDefinitionReader swaggerV1ApiDefinitionReader = new SwaggerV1ApiDefinitionReader();
+
+            ApiDefinition apiDefinition = swaggerV1ApiDefinitionReader.ReadDefinition(jobject, null);
+
+            Assert.NotNull(apiDefinition?.DirectoryStructure);
```
jimmylewis: Is this also a pre-assert?  Isn't this tested elsewhere in the overall suite?

src/Microsoft.HttpRepl/Commands/SetBaseCommand.cs
```diff
@@ -54,45 +54,29 @@ public async Task ExecuteAsync(IShellState shellState, HttpState state, ICorePar
...
             }
             else
             {
-                await SetSwaggerCommand.CreateDirectoryStructureForSwaggerEndpointAsync(shellState, state, result, cancellationToken).ConfigureAwait(false);
-                if (state.Structure != null)
-                {
-                    shellState.ConsoleManager.WriteLine(Strings.SetBaseCommand_SwaggerMetadataUriLocation + result);
-                }
-                else
+                string[] swaggerSearchPaths = new[] { "swagger.json", "swagger/v1/swagger.json" };
```
jimmylewis: should we pull these up into a static readonly field on the object?  Seems hard to find this magic list embedded this deeply in the code.

tlmii: I made this change for now, but I'm expecting this will get replaced with a setting eventually, so the lookup of the setting will be back down where that list was - that's why I left it there in the first place. But who knows when that'll happen, so... done.

src/Microsoft.HttpRepl.Tests/Commands/ChangeDirectoryCommandTests.cs
tlmii: Discussed offline. Changed it here since my concern isn't relevant for these particular block of code.

src/Microsoft.HttpRepl.Tests/Commands/SetSwaggerCommandTests.cs
tlmii: Removed.

src/Microsoft.HttpRepl.Tests/OpenApi/SwaggerV1ApiDefinitionReaderTests.cs
tlmii: It's probably more accurate to say that I was just testing two different things here (without methods in the swagger file it should still return a directory structure AND without methods in the swagger file, that directory structure should have no request info). I split it up into separate tests. I could definitely see a bug happening where when there are no methods we bail out and don't set the structure. That wasn't tested elsewhere, so I included it.

PR #164, opened 2019-08-13
Change set swagger command to allow both relative (to the base address) and absolute urls

### LINKED ISSUE #136: Getting `Must specify a swagger document`
Hey
I'm trying out httprepl and running into an issue when trying to use it right now.

This is what I'm doing:

```
dotnet httprepl https://dev.whiteboard.microsoft.com:9062/
https://dev.whiteboard.microsoft.com:9062/~ set swagger /swagger/v1.0/swagger.json
```

I then get this error: `Must specify a swagger document`

That is the relative URL to my swagger document and using Fiddler it seems like it's not even trying to make the request.

This is all that I was able to capture in Fiddler and it shows up before I use `set swagger`:

image

tlmii:
Fixes #136 by allowing the user to specify either an absolute url (which already worked) or a relative url (relative to the set base address). If no base address is set, only absolute will work, for obvious reasons.

### DIFF
--- a/src/Microsoft.HttpRepl.Fakes/MockHttpContent.cs
+++ b/src/Microsoft.HttpRepl.Fakes/MockHttpContent.cs
@@ -26,7 +26,7 @@ namespace Microsoft.HttpRepl.Fakes
 
         protected override bool TryComputeLength(out long length)
         {
-            length = Content.Length;
+            length = Content?.Length ?? 0;
             return true;
         }
     }
--- a/src/Microsoft.HttpRepl.Fakes/MockHttpContent.cs
+++ b/src/Microsoft.HttpRepl.Fakes/MockHttpContent.cs
@@ -11,22 +11,22 @@ namespace Microsoft.HttpRepl.Fakes
 {
     public class MockHttpContent : HttpContent
     {
-        public string Content { get; set; }
+        public string Content { get; }
 
         public MockHttpContent(string content)
         {
-            Content = content;
+            Content = content ?? string.Empty;
         }
 
         protected async override Task SerializeToStreamAsync(Stream stream, TransportContext context)
         {
             byte[] byteArray = Encoding.ASCII.GetBytes(Content);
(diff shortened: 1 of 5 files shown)


src/Microsoft.HttpRepl.Tests/Commands/SetSwaggerCommandTests.cs
```diff
@@ -377,12 +377,72 @@ public async Task ExecuteAsync_WithChildDirectories_SetsHttpStateSwaggerStructur
...
+    ""/api/Values"": {
+      ""post"": {
+        
+      }
+    }
+  }
+}";
+
+            string swaggerAddress = "http://localhost/swagger.json";
+            string baseAddress = "https://petstore.swagger.io/v2/";
```
jimmylewis: Can we use a junk value in the unit test?  So as to not confuse with an actual dependency on a real endpoint.

src/Microsoft.HttpRepl.Fakes/MockHttpContent.cs
```diff
@@ -26,7 +26,7 @@ protected async override Task SerializeToStreamAsync(Stream stream, TransportCon
 
         protected override bool TryComputeLength(out long length)
         {
-            length = Content.Length;
+            length = Content?.Length ?? 0;
```
jimmylewis: Is Content allowed to be null?  What's the difference between null and string.empty?
(In other words, could this also be enforced in the constructor with `Content = content ?? string.empty;`)

tlmii: Well that was an interesting rabbit hole. It looks like `Content` could sort of be null, because if `TryComputeLength` returned zero, then it wouldn't even try to call `SerializeToStreamAsync`. I'm not sure if that's guaranteed though, so we shouldn't rely on it. And making the change as you suggested, to just change `null` to `string.Empty` works the same way, but is more safe.

Issue #165, opened 2019-08-13
Notifications when preferences change


tlmii:
Currently, there's no way for the `pref set` command to notify other parts of the system that a preference has changed. 

In a lot of cases, that's ok, because the preferences are pulled on an as-needed basis. But some of them may not be. 

For example:
1. `UseDefaultCredentials`/`httpClient.useDefaultCredentials` is only checked when the `HttpClient` is first initialized. Changing the setting doesn't take effect until after a restart.
2. `HttpClientUserAgent`/`httpClient.userAgent` proposed in #XXX is only used to populate the initial set of headers, and is not checked again until after a restart.

sayedihashimi:
**Thank you for your interest in HttpRepl.**

This project has been deprecated, and the repository is now archived. As part of this transition, we are closing all open issues.
Please read the https://github.com/dotnet/HttpRepl/issues/701 for alternative recommendations.

PR #166, opened 2019-08-13
Allow the user agent header to be set via a preference

### LINKED ISSUE #123: Allow setting User-Agent header via preferences
Currently, the default value of the User-Agent header is set to `HTTP-REPL`  here:
https://github.com/aspnet/HttpRepl/blob/b208ca44a4f9361027b42df89a79594c291b72a5/src/Microsoft.HttpRepl/HttpState.cs#L45

You can override it using the various methods of setting headers, but it would be nice to have this pulled from preferences so if you, for instance, always wanted to be emulating a particular User-Agent, you could have that persist.

tlmii:
This resolves #123. 

Currently, since there's no mechanism to notify other parts of the system when a preference changes, this will require a restart. I put in #165 to address that in the future for this and other settings that it may be useful for.

### DIFF
--- a/src/Microsoft.HttpRepl.Tests/HttpStateTests.cs
+++ b/src/Microsoft.HttpRepl.Tests/HttpStateTests.cs
@@ -188,14 +188,46 @@ namespace Microsoft.HttpRepl.Tests
             Assert.Null(result);
         }
 
-        private static HttpState SetupHttpState()
+        [Fact]
+        public void HeaderSetup_WithDefaultUserAgent_UsesHttpRepl()
         {
-            IFileSystem fileSystem = new FileSystemStub();
-            IUserProfileDirectoryProvider userProfileDirectoryProvider = new UserProfileDirectoryProvider();
-            IPreferences preferences = new UserFolderPreferences(fileSystem, userProfileDirectoryProvider, null);
-            HttpClient httpClient = new HttpClient();
+            HttpState httpState = SetupHttpState(preferencesFileContent: "");
 
-            return new HttpState(fileSystem, preferences, httpClient);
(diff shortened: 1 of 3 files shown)


src/Microsoft.HttpRepl.Tests/HttpStateTests.cs
```diff
@@ -188,14 +188,46 @@ public void GetEffectivePathForPrompt_NullBaseAddress_ReturnsNull()
...
+                fileSystem = new MockedFileSystem();
+            }
+            else
+            {
+                fileSystem = new FileSystemStub();
+            }
+            UserFolderPreferences preferences = new UserFolderPreferences(fileSystem, userProfileDirectoryProvider, null);
+            if (preferencesFileContent != null)
+            {
+                ((MockedFileSystem)fileSystem).AddFile(preferences.PreferencesFilePath, preferencesFileContent);
```
jimmylewis: Could this happen in the block above?  Seems to keep the code a little simpler instead of having two checks on `preferencesFileContent`.

tlmii: It needs the preferences. Which in turn needs the filesystem. I agree its ugly.

src/Microsoft.HttpRepl/Preferences/WellKnownPreference.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using System.Reflection;
+using System.Security;
```
jimmylewis: why is this needed?

tlmii: Its not. Nor do I know why it was there. But its gone now.

PR #168, opened 2019-08-14
Adds warnings for ls/dir commands when there's no base/swagger/structure, so nothing to do.

### LINKED ISSUE #152: ls command only works when swagger doc is set.
Inform user swagger document not set when `ls` is run without document set.

tlmii:
Resolves #152

### DIFF
--- a/src/Microsoft.HttpRepl.Tests/Commands/ListCommandTests.cs
+++ b/src/Microsoft.HttpRepl.Tests/Commands/ListCommandTests.cs
@@ -8,7 +8,6 @@ using System.Threading.Tasks;
 using Microsoft.HttpRepl.Commands;
 using Microsoft.HttpRepl.Fakes;
 using Microsoft.HttpRepl.Preferences;
-using Microsoft.Repl.ConsoleHandling;
 using Microsoft.Repl.Parsing;
 using Xunit;
 
--- a/src/Microsoft.HttpRepl.Tests/Commands/ListCommandTests.cs
+++ b/src/Microsoft.HttpRepl.Tests/Commands/ListCommandTests.cs
@@ -1,10 +1,14 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.HttpRepl.Commands;
 using Microsoft.HttpRepl.Fakes;
 using Microsoft.HttpRepl.Preferences;
+using Microsoft.Repl.ConsoleHandling;
 using Microsoft.Repl.Parsing;
 using Xunit;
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.HttpRepl.Tests/Commands/ListCommandTests.cs
```diff
@@ -0,0 +1,92 @@
...
+using Microsoft.HttpRepl.Preferences;
+using Microsoft.Repl.Parsing;
+using Xunit;
+
+namespace Microsoft.HttpRepl.Tests.Commands
+{
+    public class ListCommandTests : CommandTestsBase
+    {
+        [Fact]
+        public async Task ExecuteAsync_NoBaseAddress_ShowsWarning()
```
jimmylewis: Does `..._NoBaseAddressInSwaggerFile_...` represent the scenario better?  It was a little confusing at first how it fit together until I read the other tests.

src/Microsoft.HttpRepl.Tests/Commands/ListCommandTests.cs
```diff
@@ -0,0 +1,92 @@
...
+                          baseAddress: "http://localhost/",
+                          path: "/",
+                          urlsWithResponse: null,
+                          out MockedShellState shellState,
+                          out HttpState httpState,
+                          out ICoreParseResult parseResult,
+                          out _,
+                          out IPreferences preferences);
+
+            httpState.BaseAddress = new Uri("http://localhost/swagger.json");
```
jimmylewis: set base to a swagger file in a test named NoSwagger is a little confusing.  Why isn't it just `http://localhost/`' here?  (And isn't that set by ArrangeInputs above?)

src/Microsoft.HttpRepl.Tests/Commands/ListCommandTests.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            httpState.BaseAddress = null;
+            httpState.SwaggerEndpoint = new Uri("http://localhost/swagger.json");
+            httpState.Structure = new DirectoryStructure(null);
+
+            ListCommand listCommand = new ListCommand(preferences);
+
+            await listCommand.ExecuteAsync(shellState, httpState, parseResult, CancellationToken.None);
+
+            Assert.Single(shellState.Output);
```
jimmylewis: Just something to consider: if there were multiple lines out output returned, the test would fail without telling you what they were.  I.e. you'd have to run it through the debugger to determine the failed behavior.

You could instead do something like:

```
string actualOutput = String.Join(Environment.Newline, shellState.Output);
Assert.Equal(Resources.Strings.ListCommand_Error_NoBaseAddress, actualOutput);
```

This will fail if there are more than one line (because of the Join), but will give you the full output with a single assert.

src/Microsoft.HttpRepl.Tests/Commands/ListCommandTests.cs
```diff
@@ -0,0 +1,92 @@
...
+
+namespace Microsoft.HttpRepl.Tests.Commands
+{
+    public class ListCommandTests : CommandTestsBase
+    {
+        [Fact]
+        public async Task ExecuteAsync_NoBaseAddress_ShowsWarning()
+        {
+            ArrangeInputs(commandText: "ls",
+                          baseAddress: "http://lcoalhost/",
```
jimmylewis: nit: typo.

jimmylewis: Also, why not use null here?

src/Microsoft.HttpRepl.Tests/Commands/ListCommandTests.cs
```diff
@@ -0,0 +1,92 @@
...
+using System.Threading.Tasks;
+using Microsoft.HttpRepl.Commands;
+using Microsoft.HttpRepl.Fakes;
+using Microsoft.HttpRepl.Preferences;
+using Microsoft.Repl.Parsing;
+using Xunit;
+
+namespace Microsoft.HttpRepl.Tests.Commands
+{
+    public class ListCommandTests : CommandTestsBase
```
jimmylewis: What about adding a positive test - one which doesn't show the warning?

src/Microsoft.HttpRepl/Resources/Strings.resx
```diff
@@ -264,4 +264,10 @@ When +history option is specified, commands specified in the text file will be a
     <value>If {0} is not an absolute URI, {1} must be specified.</value>
     <comment>{0} is the name of the commandSpecifiedPath parameter, {1} is the name of the baseAddress uri parameter</comment>
   </data>
-</root>
+  <data name="ListCommand_Error_NoBaseAddress" xml:space="preserve">
+    <value>No base address has been set, so there is nothing to list.</value>
```
jimmylewis: Is it clear (actionable) from these how to set the base address or directory structure?  `set base` seems like a fairly intuitive jump, but maybe we should be explicit just to be sure?

For directory structure, the action to resolve this would be to set swagger?  That might not be as intuitive from this wording.

src/Microsoft.HttpRepl.Tests/Commands/ListCommandTests.cs
tlmii: There's two conditions that cause `ls`/`dir` to not be useful - one is if there is no base address set on HttpState. The other is if there's no swagger file/structure. This test is checking the former. The next test checks the latter. More accurate would be `..._NoBaseAddressOnHttpState_...` I guess, so I'll change it to that.

src/Microsoft.HttpRepl.Tests/Commands/ListCommandTests.cs
tlmii: Yeah, that shouldn't have been there. Same with the next test. It didn't change anything (it's just checking whether BaseAddress is set to anything) but it was confusing and unnecessary, so I've removed it.

src/Microsoft.HttpRepl.Tests/Commands/ListCommandTests.cs
tlmii: Fixed.

src/Microsoft.HttpRepl/Resources/Strings.resx
tlmii: Fair. I updated both. The one for set swagger might be changing, based on my really long email and whatever comes of that discussion. But for now, this points towards set swagger.

src/Microsoft.HttpRepl.Tests/Commands/ListCommandTests.cs
tlmii: Good call. Changed.

src/Microsoft.HttpRepl.Tests/Commands/ListCommandTests.cs
tlmii: I didn't do it initially because it was a little harder to fake (the tool tries to requery the swagger url  by default, so it'll set Structure to null even if you set it to something unless you set up a response for your SwaggerEndpoint) and the integration tests already covered it. But for completeness, I went ahead and did it now.

src/Microsoft.HttpRepl.Tests/Commands/ListCommandTests.cs
```diff
@@ -12,6 +15,127 @@ namespace Microsoft.HttpRepl.Tests.Commands
...
+                          out IPreferences preferences);
+
+            httpState.SwaggerEndpoint = new Uri("http://localhost/swagger.json");
+
+            ListCommand listCommand = new ListCommand(preferences);
+
+            await listCommand.ExecuteAsync(shellState, httpState, parseResult, CancellationToken.None);
+
+            string actualOutput = string.Join(Environment.NewLine, shellState.Output);
+            Assert.DoesNotContain(Resources.Strings.ListCommand_Error_NoBaseAddress, actualOutput, StringComparison.Ordinal);
```
jimmylewis: Can we assert what the value should be, instead of multiple DoesNotContains?

PR #176, opened 2019-08-19
set bearer token

neville-nazerane:
Set a bearer token that would authorize all requests. Related to issue https://github.com/dotnet/HttpRepl/issues/156


tlmii:
Closing this PR since it won't be taken as is. We'll refer back to it when the authentication work comes back around.

PR #179, opened 2019-08-21
Initial implementation of the connect command, with tests

tlmii:
This is a second attempt at addressing #148 and cleaning up some of the syntax and auto-detection issues.

This PR creates a new command:

```
connect [rootAddress] [--base baseAddress] [--swagger swaggerAddress]
```

The argument `rootAddress` and the two options `--base` and `--swagger` are all optional, but at least one must be specified.

Some examples of how you would use this:

```console
# 1
connect https://localhost/
```
This would let the tool try to figure everything out based on a root address. It would attempt to (initially) use the root address as the base address, and then try to find the swagger address based on the root address. It would set the base address to an address specified in the swagger definition if one was present.

```console
# 2
connect --base https://localhost/
```

This is almost the same as example 1, except that the explicit base address will not be overridden by an address in the swagger definition

```console
# 3
connect --swagger https://localhost/swagger.json
```
This would skip auto-detection of the swagger address since one is explicitly specified. If no base address was present in the swagger definition, the user _could_ be left in a state where they don't have a base address.

```console
# 4
connect https://localhost/ --base v2
```
This variant explicitly sets the base address, so it won't be overridden by the swagger definition (handy if there are multiple addresses in the swagger definition). It also shows that you can use relative urls for the `--base` option.

```console
# 5
connect https://localhost/ --swagger /some/really/weird/location.json
```
This variant explicitly sets the swagger address since it is not in a well-known location. It also shows that you can use relative urls for the `--swagger` option.

```console
# 6
connect --base https://example.com/v75 --swagger https://localhost/v75/swagger.json
```
This final variant allows you to explicitly set both the base and the swagger addresses, which can be necessary if there is no obvious tie between them or if your swagger address was not in a well-known location and you wanted to pick a base address other than the first one present in the swagger definition.

~~I'll have further commits that remove the `set swagger` command and clean up the `set base` command, but wanted to get this out as soon as possible.~~ `set swagger` has been removed and `set base` has had the autodetection taken out. The former involved a little more refactoring than I realized since it was used by other commands.


src/Microsoft.HttpRepl/ApiConnection.cs
```diff
@@ -0,0 +1,153 @@
...
+            if (cancellationToken.IsCancellationRequested)
+            {
+                return null;
+            }
+
+            resp.EnsureSuccessStatusCode();
+            string responseString = await resp.Content.ReadAsStringAsync().ConfigureAwait(false);
+            JsonSerializer serializer = new JsonSerializer { PreserveReferencesHandling = PreserveReferencesHandling.All };
+            JObject responseObject = (JObject)serializer.Deserialize(new StringReader(responseString), typeof(JObject));
+            ApiDefinitionReader reader = new ApiDefinitionReader();
```
jodavis: `reader` isn't used

tlmii: Removed.

PR #201, opened 2019-08-30
Fix BaseHttpCommand to only print the request body once if echo is on

### LINKED ISSUE #196: Echoing requests prints out the request body twice
One time it comes through colored, one time it doesn't.

tlmii:
Fixes #196.

For some reason, there was an extra loop outputting the request body after the response marker. This removes that.

### DIFF
--- a/src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
+++ b/src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
@@ -380,11 +380,6 @@ namespace Microsoft.HttpRepl.Commands
             consoleManager.WriteLine();
             consoleManager.WriteLine($"Response from {hostString}...".SetColor(requestConfig.AddressColor));
             consoleManager.WriteLine();
-
-            foreach (string responseLine in responseOutput)
-            {
-                consoleManager.WriteLine(responseLine);
-            }
         }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
```diff
@@ -380,11 +380,6 @@ private async Task HandleResponseAsync(HttpState programState, DefaultCommandInp
             consoleManager.WriteLine();
             consoleManager.WriteLine($"Response from {hostString}...".SetColor(requestConfig.AddressColor));
```
jimmylewis: It's a little bit weird that this ends with a header for the response content (which it no longer prints), but since it's only called from one place I see how it pieces together.  It might flow better if this part were moved out to `HandleResponseAsync` after it calls `HandleEchoRequest`.  Up to you though, I don't have a strong preference either way.

tlmii: Agreed. I went back and forth on this several times. It was weird to find the head for the response in the HandleEcho section... but it was also kind of appropriate because it was only written out if you had to handle echo...  I went ahead and moved it out to the outer call, inside the if block. Had to promote some variables and pass them as parameters too. I _think_ it's cleaner, but I'm not really sure :)

PR #208, opened 2019-09-02
Set the temp file extension based on the content type

### LINKED ISSUE #186: File extension on temp file for body
Great project,

as an idea, could you add an appropriate extension to the temp file name, so that an editor can provide colouring and intellisense for the body template?

Should be here:
https://github.com/dotnet/HttpRepl/blob/0189a224b07fa1b82b3cedf36886ede778e131a9/src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs#L202

Thanks!

tlmii:
Resolves #186. 

When the content type we are sending is something known (`application/json`, `application/xml`, `text/xml` currently), we will create a temp file name with the appropriate file extension. If its anything else, we just use the default `Path.GetTempFileName()`.


src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
```diff
@@ -37,6 +37,12 @@ public abstract class BaseHttpCommand : CommandWithStructuredInputBase<HttpState
         private const string StreamingOption = nameof(StreamingOption);
         private const string BodyContentOption = nameof(BodyContentOption);
         private static readonly char[] HeaderSeparatorChars = new[] { '=', ':' };
+        private static readonly Dictionary<string, string> FileExtensionLookup = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
+        {
+            { "application/json", ".json" },
```
jimmylewis: technically, text/json also exists, though it's not preferred or as widespread.

src/Microsoft.HttpRepl/Properties/AssemblyInfo.cs
```diff
@@ -1,3 +1,4 @@
 using System.Runtime.CompilerServices;
 
+[assembly: InternalsVisibleTo("Microsoft.HttpRepl.IntegrationTests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
```
jimmylewis: :( I dislike opening an avenue to integration tests being too tightly coupled to the implementation details.  Can we just make RealFileSystem public? (I assume that's the type this was added for)

src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
tlmii: added.

src/Microsoft.HttpRepl/FileSystem/RealFileSystem.cs
```diff
@@ -40,9 +42,31 @@ public void DeleteFile(string path)
...
 
-        public string GetTempFileName()
+        public string GetTempFileName(string fileExtension = null)
         {
-            return Path.GetTempFileName();
+            // If they don't care about the extension, go the easy route and use the
+            // system-provided way of creating a unique temp file.
+            if (fileExtension is null)
+            {
+                return Path.GetTempFileName();
```
jimmylewis: There are now two fairly different behaviors of this method:

1. Called with no extension, it will create a 0-byte file on disk and return the path to it (what `Path.GetTempFileName()` does).  The new file is guaranteed to be unique as long as the system can provide one.
2. Called with an extension, it will return a string to a filename easily associated with http repl; however, no file is created and no check is made for duplicates (which is rare, but possible).

It feels like there's a design inconsistency here:
* should we always create the file?
* should we guarantee name uniqueness?
* should we always use the `HttpRepl.{Guid}` name, and use .tmp as a default extension?

src/Microsoft.HttpRepl/Properties/AssemblyInfo.cs
tlmii: Yep, that was it. My instinct to make things as restricted as possible tends to conflict with my instinct to have fewer IVTs. Swapped which one won out this time.

src/Microsoft.HttpRepl/Resources/Strings.resx
```diff
@@ -350,4 +350,8 @@ When +history option is specified, commands specified in the text file will be a
   <data name="SetBaseCommand_MustSpecifyServerError" xml:space="preserve">
     <value>Must specify a server</value>
   </data>
+  <data name="RealFileSystem_Error_InvalidExtension" xml:space="preserve">
+    <value>If specified, {0} must begin with a period.</value>
```
jimmylewis: Is this an actionable user message?  This seems more like a developer-oriented message.

src/Microsoft.HttpRepl.IntegrationTests/FileSystem/RealFileSystemTests.cs
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using Microsoft.HttpRepl.FileSystem;
+using Xunit;
+
+namespace Microsoft.HttpRepl.IntegrationTests.FileSystem
```
jimmylewis: These feel like unit tests more than integration tests, even though they have an concrete dependency that can't be faked.  Thoughts?

src/Microsoft.HttpRepl/FileSystem/RealFileSystem.cs
tlmii: As far as the uniqueness goes, I was basing my "good enough" call on Jared's answer here on a similarly constructed temp file name: https://stackoverflow.com/a/581588/228298.
> Note, it's not guaranteed to produce a unique file name since someone else could have technically already created that file. However the chances of someone guessing the next guid produced by your app and creating it is very very low. It's pretty safe to assume this will be unique.

As far as whether the file gets created or not... I'd prefer it NOT get created... but there's no way to turn that off while still using `GetTempFileName()`, which brings me to the last point...

>should we always use the HttpRepl.{Guid}` name and use .tmp as a default extension?

Probably. I think that's the way I'm leaning. It gets rid of the unnecessarily created 0-byte file, and it keeps the consistency of being able to locate the file more easily.

If you're fine with the answer on uniqueness above, then I'll go that route.

src/Microsoft.HttpRepl/Resources/Strings.resx
tlmii: You're correct. Your point being, I assume, that it doesn't need to be in `Strings`.

src/Microsoft.HttpRepl.IntegrationTests/FileSystem/RealFileSystemTests.cs
tlmii: I went back and forth on that a bit. I would have preferred to make them unit tests, but the reliance on `System.IO.Path.GetTempFileName()` (primarily) and `System.IO.Path.GetTempPath()`  (somewhat) pushed me towards Integration Tests since we wouldn't want to be doing anything with the file system in unit tests.

With one of the changes discussed above, we'd be getting rid of `GetTempFileName()` usage, so that's a positive in the move towards unit tests. But we still have a reliance on `GetTempPath()`. 

So then I guess the question becomes "is it worth it to create a separate `ITempFileSystemProvider` (terrible name, but you get the point) to use here for that one function? 

I'd lean towards no - for the moment at least - in part because it seems reasonable that things dealing with the actual file system (as these particular tests do) are in integration tests.

Thoughts?

src/Microsoft.HttpRepl/FileSystem/RealFileSystem.cs
jimmylewis: A uniqueness check could be done with a retry loop, but I agree that it's highly unlikely to happen, especially given the HttpRepl prefix.  Making the behavior consistent is the main change.

src/Microsoft.HttpRepl.IntegrationTests/FileSystem/RealFileSystemTests.cs
jimmylewis: To me, an integration test is a test that drives the product and focuses on a user scenario rather than the correctness of a single component.  With the removal of `GetTempFileName()` the code being tested here no longer even has side effects on disk; while that's technically an implementation detail, it removes the biggest concern towards low-impact unit tests.  Yes, it does still rely on `Path.GetTempPath()`, but you've already shown that it's easy to test around that, and I don't have concerns about a dependency on a platform-variant behavior since it doesn't have external side effects.

I don't think it requires a further `ITempFileSystemProvider` level of abstraction.  That seems overly complicated at this time.

tlmii: I see what you're getting at. Shifting them to unit tests as-is "just works", so I've done that.

src/Microsoft.HttpRepl.Tests/FileSystem/RealFileSystemTests.cs
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using Microsoft.HttpRepl.FileSystem;
+using Xunit;
+
+namespace Microsoft.HttpRepl.IntegrationTests.FileSystem
```
jodavis: Now that this is a unit test, this should be "Microsoft.HttpRepl.**Tests**.FileSystem"

Also, since the unit test assembly already has IVT access, can RealFileSystem be made internal again?

tlmii: Good catch. Changed.

src/Microsoft.HttpRepl.Tests/FileSystem/RealFileSystemTests.cs
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Microsoft.HttpRepl.Tests.FileSystem
+{
+    public class RealFileSystemTests
+    {
+        [Theory]
+        [InlineData(".json")]
+        [InlineData(".xml")]
+        [InlineData(".tmp")]
+        public void GetTempFileName_WithExtension_ReturnsFileNameWithExtension(string extension)
```
jimmylewis: Should this have a WithoutExtension test too?

src/Microsoft.HttpRepl/FileSystem/RealFileSystem.cs
```diff
@@ -40,9 +42,28 @@ public void DeleteFile(string path)
...
-        public string GetTempFileName()
+        public string GetTempFileName(string fileExtension)
         {
-            return Path.GetTempFileName();
+            if (fileExtension is null)
+            {
+                throw new ArgumentNullException(nameof(fileExtension));
+            }
+
+            if (!fileExtension.StartsWith(".", StringComparison.Ordinal))
```
jimmylewis: What should it do if `fileExtension == "."`?  Is that a valid input?

src/Microsoft.HttpRepl.Tests/FileSystem/RealFileSystemTests.cs
jimmylewis: never mind, I see you removed that overload.  Does WithExtension need to be in the test name?

src/Microsoft.HttpRepl.Tests/FileSystem/RealFileSystemTests.cs
```diff
@@ -0,0 +1,56 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using Microsoft.HttpRepl.FileSystem;
+using Xunit;
+
+namespace Microsoft.HttpRepl.Tests.FileSystem
+{
+    public class RealFileSystemTests
```
jimmylewis: Add tests for the validation failure cases?
- null
- extension doesn't start with "."
- extension is only "." (assuming that's invalid... well, you could test it either way)

src/Microsoft.HttpRepl.Tests/FileSystem/RealFileSystemTests.cs
tlmii: I changed it to WithValidInput, since I also created WithInvalidInput tests. That seemed better than just removing it.

src/Microsoft.HttpRepl/FileSystem/RealFileSystem.cs
tlmii: As discussed, changed it to disallow that.

src/Microsoft.HttpRepl.Tests/FileSystem/RealFileSystemTests.cs
tlmii: Added more tests for those scenarios.

PR #209, opened 2019-09-02
Rework ui command to be more flexible

### LINKED ISSUE #151: ui command assumes /swagger endpoint
`ui             Displays the Swagger UI page, if available, in the default browser`

Create an option to pass or set the endpoint for the swagger ui

tlmii:
Resolves #151.

Previously, the `ui` command just tried to go to a url formed (essentially) from:

```C#
Uri uri = new Uri(baseAddress, swagger);
```

This was limited, for obvious reasons.

This PR adds flexibility in two ways:

1. The `ui` command now takes an optional parameter, where you can specify the absolute or relative (to the base address) swagger endpoint. So, for instance, if your base address is https://localhost:44366/ and your swagger endpoint is https://localhost:44366/mySwaggerEndpoint, you can issue either `ui mySwaggerEndpoint` or `ui https://localhost:44366/mySwaggerEndpoint` and it should launch the correct ui for you.
2. The `ui` command, absent a provided parameter, will attempt to use the `swagger.uiEndpoint` preference, if set, before falling back to the original behavior of trying `swagger`. So you could call `pref set swagger.uiEndpoint mySwaggerEndpoint` and then all subsequent calls to ui (including both in the current session and subsequent sessions) would use that value (with your current base address, since it is a relative value).

### DIFF
--- a/src/Microsoft.HttpRepl/Preferences/WellKnownPreference.cs
+++ b/src/Microsoft.HttpRepl/Preferences/WellKnownPreference.cs
@@ -172,6 +172,8 @@ namespace Microsoft.HttpRepl.Preferences
 
         public static string SwaggerSearchPaths { get; } = "swagger.searchPaths";
 
+        public static string SwaggerUIEndpoint { get; } = "swagger.uiEndpoint";
+
         public static string UseDefaultCredentials { get; } = "httpClient.useDefaultCredentials";
 
         public static string HttpClientUserAgent { get; } = "httpClient.userAgent";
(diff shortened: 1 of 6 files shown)


src/Microsoft.HttpRepl/Preferences/WellKnownPreference.cs
```diff
@@ -172,6 +172,8 @@ public static IReadOnlyList<string> Names
 
         public static string SwaggerSearchPaths { get; } = "swagger.searchPaths";
 
+        public static string SwaggerUIEndpoint { get; } = "swagger.uiEndpoint";
```
jimmylewis: Is this appropriate as a user setting?  It seems like it would vary from project to project, even though conventions may bring some level of consistency...

tlmii: Discussed offline. Going to leave it as a setting to provide a balance between having to pass it as a parameter to the `ui` command every time and having it always use the `swagger` relative uri. Would be nice if it was specified somewhere in the swagger/openapi definition, but it is not.

PR #225, opened 2019-09-10
Adding in-app documentation for the ui command

tlmii:
When we pulled #209 we forgot to include in-app documentation on the new flexibility of the `ui` command. This PR adds that documentation. 

The output will now be:

```console
Usage: ui [{swaggerUIAddress}]                                                                                          
Launches the Swagger UI page (if available) in the default browser

The ui command uses multiple sources to determine the swagger ui endpoint. In order of precedence:
  1. The {swaggerUIAddress} parameter, if specified
  2. The swagger.uiEndpoint preference, if set
  3. A default url of [BaseAddress]/swagger
```

@scottaddie Thoughts on this? I figure this can be the basis for the changes to the docs once we settle on it here.

### DIFF
--- a/src/Microsoft.HttpRepl/Resources/Strings.resx
+++ b/src/Microsoft.HttpRepl/Resources/Strings.resx
@@ -220,7 +220,7 @@
     <value>Must specify a swagger document</value>
   </data>
   <data name="UICommand_Description" xml:space="preserve">
-    <value>ui - Launches the Swagger UI page (if available) in the default browser</value>
+    <value>Launches the Swagger UI page (if available) in the default browser</value>
   </data>
   <data name="UICommand_NotConnectedToServerError" xml:space="preserve">
     <value>Must be connected to a server to launch Swagger UI</value>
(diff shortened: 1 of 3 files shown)


src/Microsoft.HttpRepl/Resources/Strings.resx
```diff
@@ -358,4 +358,20 @@ When +history option is specified, commands specified in the text file will be a
     <value>The parameter '{0}' could not be converted into a valid uri.</value>
     <comment>{0} is a string parameter to the UI command</comment>
   </data>
+  <data name="UICommand_HelpText_Line1" xml:space="preserve">
+    <value>The {0} command uses multiple sources to determine the swagger ui endpoint. In order of precedence:</value>
```
scottaddie:
```suggestion
    <value>The {0} command uses multiple sources to determine the Swagger UI endpoint. In order of precedence:</value>
```

src/Microsoft.HttpRepl/Resources/Strings.resx
```diff
@@ -358,4 +358,20 @@ When +history option is specified, commands specified in the text file will be a
...
+  <data name="UICommand_HelpText_Line2" xml:space="preserve">
+    <value>  1. The {0} parameter, if specified</value>
+    <comment>{0} is the syntax for the parameter, e.g. {swaggerUIAddress}</comment>
+  </data>
+  <data name="UICommand_HelpText_Line3" xml:space="preserve">
+    <value>  2. The {0} preference, if set</value>
+    <comment>{0} is the name of the preference, e.g. swagger.uiEndpoint</comment>
+  </data>
+  <data name="UICommand_HelpText_Line4" xml:space="preserve">
+    <value>  3. A default url of {0}</value>
```
scottaddie:
```suggestion
    <value>  3. A default URL of {0}</value>
```

PR #249, opened 2019-09-29
Adding Analyzers

tlmii:
This PR is all about enabling analyzers on HttpRepl and addressing the errors that came from it. I took the `.ruleset` files from https://github.com/aspnet/LibraryManager. This PR touched a good number of files, so I broke up the changes into four parts:

1. Commit 022d9da split the folder structure off so that test projects were in their own project separate from the main source. 
2. Commit f520782 enabled analyzers for the src/ projects
3. Commit fe6b193 addressed all of the errors that were not related to strings/resources
4. Commit 09d3d78 addressed the string/resources errors

For that last commit, the analzers wanted me to put things like `HTTP`, `;`, `,`, `:`, and `/` into resources. I realized it was because the `SetColor` exension method's first parameter was named `text`. Renaming that to `textToColor` made the errors go away. I addressed all the ones that were actual localizable text first before doing that..


src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
```diff
@@ -557,21 +581,38 @@ protected override string GetHelpDetails(IShellState shellState, HttpState progr
...
         {
+            if (programState is null)
+            {
+                throw new ArgumentNullException(nameof(programState));
+            }
+
             List<string> results = new List<string>();
 
-            if (programState.Structure != null && programState.BaseAddress != null)
+            if (!(programState.Structure is null) && !(programState.BaseAddress is null))
```
jimmylewis: Where did we land on the `foo is object` vs `foo != null` conversation?

src/Microsoft.HttpRepl/Commands/TreeNode.cs
```diff
@@ -15,7 +15,9 @@ public class TreeNode
         private readonly List<TreeNode> _children = new List<TreeNode>();
 
         public TreeNode(Formatter formatter, string prefix, string entry)
-            : this(formatter, prefix, entry, 0)
+            : this(formatter ?? throw new ArgumentNullException(nameof(formatter)),
```
jimmylewis: Seems like the null check should be a concern for the other constructor.  Even though it's private, if an e.g. factory method were added, it ought to have the same checks applied.

src/Microsoft.HttpRepl/Commands/TreeNode.cs
```diff
@@ -30,7 +32,7 @@ private TreeNode(Formatter formatter, string prefix, string entry, int depth)
 
         public TreeNode AddChild(string prefix, string entry)
         {
-            TreeNode child = new TreeNode(_formatter, prefix, entry, _depth + 1);
+            TreeNode child = new TreeNode(_formatter, prefix ?? throw new ArgumentNullException(nameof(prefix)), entry, _depth + 1);
```
jimmylewis: it feels weird to embed the null check in the constructor arguments.  For readability, I'd've expected the null check first, then the call to the constructor.  Also, in tandem with the other comment here, this null check seems like it belongs inside the private constructor, not inline in the call.

src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
```diff
@@ -92,6 +94,21 @@ public override CommandInputSpecification InputSpec
 
         protected override async Task ExecuteAsync(IShellState shellState, HttpState programState, DefaultCommandInput<ICoreParseResult> commandInput, ICoreParseResult parseResult, CancellationToken cancellationToken)
         {
+            if (programState is null)
```
jimmylewis: This 4-line pattern comes up a lot, how do you feel about the more concise `programState = programState ?? throw new ...`?

src/Microsoft.HttpRepl/Microsoft.HttpRepl.csproj
```diff
@@ -20,7 +20,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\Microsoft.Repl\Microsoft.Repl.csproj" />
+    <ProjectReference Include="$(RepoSource)Microsoft.Repl\Microsoft.Repl.csproj" />
```
jimmylewis: I find it much more readable to have the `\` separator after variables.  Can we keep the slash off in `$(RepoSource)` so it can be put here instead?

src/Microsoft.HttpRepl/Properties/AssemblyInfo.cs
```diff
@@ -1,3 +1,8 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
 
 [assembly: InternalsVisibleTo("Microsoft.HttpRepl.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
+[assembly: SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "This is done commonly throughout the codebase to catch unexpected errors.") ]
```
jimmylewis: Why not suppress this in the .ruleset file?  Or do you only want the suppression to apply to this project?

tlmii: Yeah, that's exactly it. I didn't even really want to put it at the assembly level, but it was on too many files to really avoid it.

src/Microsoft.HttpRepl/Microsoft.HttpRepl.csproj
tlmii: I was trying to make it consistent with `$(RepoRoot)` which I can't change (Arcade uses it) and which has the slash. I agree about the readability though. I changed it, but I still don't like the inconsistency.

src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
tlmii: I've never really been a huge fan of it, since in the common case it means (from a code read perspective, not an execution perspective, I assume) you've got an extra self-assignment that you have to ignore. But I agree those four-line blocks were repetitive and annoying. I changed it.

src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
tlmii: Not sure where we landed. But I went ahead and changed it for this repo. I was tempted to use the newer `if (foo is { })` but that required upgrading to C# 8 and I didn't want to make that change in this PR :)

src/Microsoft.HttpRepl/Commands/TreeNode.cs
tlmii: I thought I had tried that and it was still throwing the Analyzer error on the public constructor. But I just tried it again and it seems to be ok with it. Maybe I hadn't fully refreshed the analyzer results (sometimes VS seems to not clear the list).

src/Microsoft.HttpRepl/Commands/TreeNode.cs
tlmii: Same answer as above.

src/Microsoft.HttpRepl/Commands/TreeNode.cs
```diff
@@ -21,9 +21,9 @@ public TreeNode(Formatter formatter, string prefix, string entry)
 
         private TreeNode(Formatter formatter, string prefix, string entry, int depth)
         {
-            _formatter = formatter;
+            _formatter = formatter ?? throw new ArgumentNullException(nameof(formatter));
             formatter.RegisterEntry(prefix.Length, depth);
```
jimmylewis: You haven't checked prefix for null before using `prefix.Length` here.

(also implied: there's a missing unit test)

src/Microsoft.HttpRepl/DirectoryStructureExtensions.cs
```diff
@@ -15,12 +16,16 @@ public static IEnumerable<string> GetDirectoryListingAtPath(this IDirectoryStruc
 
         public static IDirectoryStructure TraverseTo(this IDirectoryStructure structure, string path)
         {
+            path = path ?? throw new ArgumentNullException(nameof(path));
+
             string[] parts = path.Replace('\\', '/').Split('/');
             return structure.TraverseTo(parts);
```
jimmylewis: do you need to check `structure` for null too?  It's checked in the next overload (and that one doesn't check `pathParts` for null...)

src/Microsoft.HttpRepl/Formatting/JsonVisitor.cs
```diff
@@ -11,104 +12,109 @@ namespace Microsoft.HttpRepl.Formatting
 {
     public static class JsonVisitor
     {
+#pragma warning disable CA1308 // Normalize strings to uppercase
```
jimmylewis: nit: I'm not a big fan of looong #pragma regions, it's easy to refactor and not realize the pragma was in effect.  Worth just suppressing with attribute on this method?

src/Microsoft.HttpRepl/Program.cs
```diff
@@ -127,7 +130,9 @@ private static HttpClient GetHttpClientWithPreferences(IPreferences preferences)
         {
             if (preferences.GetBoolValue(WellKnownPreference.UseDefaultCredentials))
             {
+#pragma warning disable CA2000 // Dispose objects before losing scope
                 return new HttpClient(new HttpClientHandler { UseDefaultCredentials = true });
```
jimmylewis: Who owns disposal of the HttpClient?  How often are we creating new instances?

tlmii: There should be exactly one `HttpClient` for the lifetime of the application. We don't dispose it directly, but we'd only do so on shutdown of the application anyway. For the record, that CA2000 error was for HttpClientHandler, not HttpClient.

src/Microsoft.HttpRepl/Commands/TreeNode.cs
tlmii: Checked, added tests and re-ordered the checks too.

src/Microsoft.HttpRepl/DirectoryStructureExtensions.cs
tlmii: Yep. Done.

src/Microsoft.HttpRepl/Formatting/JsonVisitor.cs
tlmii: Yeah, I should have done it that way to begin with. Done.

Issue #237, opened 2019-09-20
POST/PATCH with body adds custom headers twice


djmitchella:
If I do `echo on`, `cd /api/things`, `set header my-header abcd`, and then do:
````
GET
````
I can see `my-header: abcd` showing up once in the list of headers.

But if I try a POST with a body:
````
POST -c "hello"
````
then I get `my-header: abcd` twice in the list of headers, and if the API I'm talking to doesn't like duplicated headers, I get problems. If I don't supply a body, I don't get the duplicated header, it only shows up once.

Same problem happens with PATCH; if I supply a body, then my custom header gets sent twice.

tlmii:
It looks like the headers are _always_ getting added to the request's headers collection, and then, if there is a body, they're _also_ getting added to the content's headers collection. Thus the duplication. Thanks for the bug report. I'll see about getting a PR out to fix this.

PR #277, opened 2019-11-01
update to fix duplicated ContentType with Test

rposener:
Fix to #237  provided with UnitTest coverage.


tlmii:
I just completed #354, which goes a little bit further in addressing the issue. Thanks for taking the time to submit the PR and nudge this issue forward!

Issue #336, opened 2020-03-27
Improve initial header output when repl starts


timheuer:
When the repl tool is started we should emit some more helpful information on how to use it, at a minimum a link to the docs of where to get more information, e.g., an aka.ms link to helpful information. - use https://aka.ms/httprepl

bradygaster:
@timheuer - what's the status of the pull request on this? We'd like to get this one wrapped up - let us know next steps as you see them

tlmii:
Completed #337, so closing this.

PR #337, opened 2020-03-30
Fix for #336 - Emit the helper doc link on connect

timheuer:
Based on some customer feedback, when users launch HttpRepl for the first time, there is no guidance on the commands.  This attempts to address this by using the existing help command pointing to the docs upon connect.

### DIFF
--- a/src/Microsoft.HttpRepl/Commands/ConnectCommand.cs
+++ b/src/Microsoft.HttpRepl/Commands/ConnectCommand.cs
@@ -111,6 +111,7 @@ namespace Microsoft.HttpRepl.Commands
                 shellState.ConsoleManager.WriteLine(string.Format(Resources.Strings.ConnectCommand_Status_Swagger, programState.SwaggerEndpoint));
             }
 
+            // Always show help link after connecting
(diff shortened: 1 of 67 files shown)


timheuer:
CR @tlmii ?

src/Microsoft.HttpRepl/Commands/ConnectCommand.cs
```diff
@@ -109,6 +109,8 @@ private static void WriteStatus(IShellState shellState, HttpState programState)
             {
                 shellState.ConsoleManager.WriteLine(string.Format(Resources.Strings.ConnectCommand_Status_Swagger, programState.SwaggerEndpoint));
             }
+
+            shellState.ConsoleManager.WriteLine(Resources.Strings.HelpCommand_Core_Details_Line2.Bold().Cyan());
```
tlmii, line 113: Since the resource string name isn't very helpful at identifying what's being written, I'd add a comment saying what we're always writing to the console.

```suggestion
            // Always show help link after connecting
            shellState.ConsoleManager.WriteLine(Resources.Strings.HelpCommand_Core_Details_Line2.Bold().Cyan());
```

Issue #346, opened 2020-05-25
Rename --swagger option to --openapi


darrelmiller:
https://github.com/dotnet/HttpRepl/blob/85045582b5bd35ba8a09fd44cc1a05ee8e31472b/src/Microsoft.HttpRepl/Commands/ConnectCommand.cs#L39

The API description document is called an OpenAPI description.  Swagger is a trademarked term owned by Smartbear that identifies a set of tools for working with OpenAPI descriptions.

tlmii:
Guessing we will probably want to add `--openapi` as an alternative, leaving `--swagger` in place initially (but changing all the documentation references) and then we can cut `--swagger` eventually. 

Could possibly add a warning saying --swagger was deprecated as well. 

I'll take a look.

bradygaster:
agreed on this - at the very least let's make this an alias.

tlmii:
Resolved by #353

Issue #248, opened 2019-09-29
Search for OpenAPI v3 filename


Tasteful:
Regarding to [OpenAPI.v3](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#document-structure) the recommendation is that the spec name should be `openapi.json` or `openapi.yaml`. Should that be added to the searchpath's below?

https://github.com/dotnet/HttpRepl/blob/53e84a988a96d2d959801f8205a8babdef56b873/src/Microsoft.HttpRepl/ApiConnection.cs#L21

tlmii:
Good call. You're right it probably should be.

PR #353, opened 2020-06-23
Swagger Cleanup

tlmii:
Three small commits in this PR:

1. Cleans up some leftover references to the old `set swagger` command that no longer existed when we introduced the `connect` command.
2. Adds `--openapi` and `-o` as the preferred ways of specifying the OpenAPI description document. This addresses #346. `--swagger` is still left as an option, but this removes it from (most of) the documentation. We can probably deprecate it when we get to 4.0.
3. Adds `openapi.json` and `/openapi.json` to the search paths for auto discovery of the OpenAPI description url. THis addresses #248. 

cc @darrelmiller since I can't seem to add you as a reviewer.

### DIFF
--- a/src/Microsoft.HttpRepl/ApiConnection.cs
+++ b/src/Microsoft.HttpRepl/ApiConnection.cs
@@ -18,7 +18,7 @@ namespace Microsoft.HttpRepl
     internal class ApiConnection
     {
         private readonly IPreferences _preferences;
-        private const string SwaggerSearchPaths = "swagger.json|swagger/v1/swagger.json|/swagger.json|/swagger/v1/swagger.json";
+        private const string SwaggerSearchPaths = "swagger.json|swagger/v1/swagger.json|openapi.json|/swagger.json|/swagger/v1/swagger.json|/openapi.json";
 
         public Uri RootUri { get; set; }
         public bool HasRootUri => RootUri is object;
--- a/src/Microsoft.HttpRepl/Commands/ConnectCommand.cs
+++ b/src/Microsoft.HttpRepl/Commands/ConnectCommand.cs
@@ -40,7 +40,8 @@ namespace Microsoft.HttpRepl.Commands
                                                                                                                                    requiresValue: true,
                                                                                                                                    minimumOccurrences: 0,
                                                                                                                                    maximumOccurrences: 1,
-                                                                                                                                   forms: new[] { "--swagger", "-s" }))
+                                                                                                                                   forms: new[] { "--openapi", "-o",
+                                                                                                                                                  "--swagger", "-s" }))
                                                                                         .Finish();
 
         public override string GetHelpSummary(IShellState shellState, HttpState programState)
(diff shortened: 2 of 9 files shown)


src/Microsoft.HttpRepl/Resources/Strings.resx
```diff
@@ -121,7 +121,7 @@
     <value>Append the given directory to the currently selected path, or move up a path when using `cd ..`</value>
   </data>
   <data name="ChangeDirectoryCommand_Warning_UnknownEndpoint" xml:space="preserve">
-    <value>Warning: The '{0}' endpoint is not present in the Swagger metadata</value>
+    <value>Warning: The '{0}' endpoint is not present in the OpenAPI description</value>
```
bradygaster, line 124: 👍

src/Microsoft.HttpRepl/Commands/ConnectCommand.cs
```diff
@@ -54,7 +55,7 @@ protected override string GetHelpDetails(IShellState shellState, HttpState progr
             {
                 var helpText = new StringBuilder();
                 helpText.Append(Resources.Strings.Usage.Bold());
-                helpText.AppendLine("connect [rootAddress] [--base baseAddress] [--swagger swaggerAddress]");
+                helpText.AppendLine("connect [rootAddress] [--base baseAddress] [--openapi openApiDescriptionAddress]");
```
bradygaster, line 58: 👍

src/Microsoft.HttpRepl/Resources/Strings.resx
```diff
@@ -286,23 +274,23 @@ When +history option is specified, commands specified in the text file will be a
     <value>Unable to determine a base address</value>
   </data>
   <data name="ConnectCommand_Status_Swagger" xml:space="preserve">
```
bzhu94, line 276: Change the name to "ConnectCommand_Status_OpenAPI" or something similar?

src/Microsoft.HttpRepl/ApiConnection.cs
```diff
@@ -18,7 +18,7 @@ namespace Microsoft.HttpRepl
     internal class ApiConnection
     {
         private readonly IPreferences _preferences;
-        private const string SwaggerSearchPaths = "swagger.json|swagger/v1/swagger.json|/swagger.json|/swagger/v1/swagger.json";
+        private const string SwaggerSearchPaths = "swagger.json|swagger/v1/swagger.json|openapi.json|/swagger.json|/swagger/v1/swagger.json|/openapi.json";
```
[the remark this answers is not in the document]
tlmii: If its common to use those paths without the extension, then we should probably add it. I will note that the OAS spec suggests paths _with_ the extension though, so that's what I started with here. Can always add more later.

src/Microsoft.HttpRepl/Resources/Strings.resx
tlmii: I'm going to do a second pass at this later on to clean up the non-user-facing uses of swagger as appropriate.

src/Microsoft.HttpRepl/Resources/Strings.resx
```diff
@@ -286,23 +274,23 @@ When +history option is specified, commands specified in the text file will be a
...
-    <value>Using swagger definition at {0}</value>
+    <value>Using OpenAPI description at {0}</value>
     <comment>{0} indicates a uri</comment>
   </data>
   <data name="ConnectCommand_Status_NoSwagger" xml:space="preserve">
```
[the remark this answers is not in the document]
tlmii: I'm going to do a second pass at this later on to clean up the non-user-facing uses of swagger as appropriate.

src/Microsoft.HttpRepl/ApiConnection.cs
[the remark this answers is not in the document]
spboyer: is `oas` a known term or is this something we are causally using and/or inventing?

PR #354, opened 2020-06-25
Don't duplicate headers

### LINKED ISSUE #237: POST/PATCH with body adds custom headers twice

tlmii:
Fixes #237.

There were actually two separate things going on:

1. As pointed out by #237, almost all headers were duplicated if you were doing anything with a request body (because headers were being set for both the request and the content).
2. The fix in #277 showed the other part of it - that Content-Type specifically was always being duplicated (with or without a request body) if it was added with `set header` or the `--header` argument, because it was special-cased and set early before all those were added in bulk.

This PR addresses both issues, and supersedes #277, which I'll close with an explanation once this goes in.

### DIFF
--- a/src/Microsoft.HttpRepl/WellKnownHeaders.cs
+++ b/src/Microsoft.HttpRepl/WellKnownHeaders.cs
@@ -1,17 +1,14 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.Collections.Generic;
-using System.Diagnostics;
-using System.Runtime.InteropServices;
-using System.Threading.Tasks;
-using Microsoft.HttpRepl.Resources;
 
 namespace Microsoft.HttpRepl
 {
     public static class WellKnownHeaders
     {
+        public static readonly string ContentType = "Content-Type";
+
         public static readonly IEnumerable<string> CommonHeaders = new[]
         {
             "A-IM",
@@ -91,7 +88,5 @@ namespace Microsoft.HttpRepl
 
             "Last-Modified",
         };
-
-        public static readonly string ContentType = "Content-Type";
     }
 }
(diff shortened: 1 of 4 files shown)


src/Microsoft.HttpRepl/WellKnownHeaders.cs
```diff
@@ -0,0 +1,92 @@
...
+            "Content-Disposition",
+            "Content-Encoding",
+            "Content-Language",
+            "Content-Length",
+            "Content-Location",
+            "Content-MD5",
+            "Content-Range",
+            ContentType,
+            "Expires",
+
```
bzhu94, line 88: Nit: excess whitespace

test/Microsoft.HttpRepl.Tests/Commands/PostCommandTests.cs
```diff
@@ -77,6 +78,87 @@ public async Task ExecuteAsync_OnlyBaseAddressWithInlineContent_VerifyResponse()
...
+                out HttpState httpState,
+                out ICoreParseResult parseResult,
+                out MockedFileSystem fileSystem,
+                out IPreferences preferences);
+
+            httpState.EchoRequest = true;
+
+            PostCommand postCommand = new PostCommand(fileSystem, preferences);
+            await postCommand.ExecuteAsync(shellState, httpState, parseResult, CancellationToken.None);
+
```
bzhu94, line 99: Nit: excess whitespace

test/Microsoft.HttpRepl.Tests/Commands/PostCommandTests.cs
```diff
@@ -77,6 +78,87 @@ public async Task ExecuteAsync_OnlyBaseAddressWithInlineContent_VerifyResponse()
...
+
```
bzhu94, line 127: Nit: excess whitespace

PR #358, opened 2020-06-29
Address security vulnerability

tlmii:
Security vulnerability scanning turned up a reference that needed to be updated. This PR fixes that.


tlmii:
@jimmylewis After we get 3.1 out the door I think we're going to have another larger release (w/ the OpenAPI.NET and System.CommandLine changes) that'll have a lot longer bake time. I'll try switching it during that timeframe.

PR #366, opened 2020-07-30
Fix issue with extra spaces in commands

### LINKED ISSUE #365: Extra space in set base command leads to error
I copied a url from the web, and pasted that into my `set base` command. I accidently copied a leading space in the url which I did not notice. The command is below.

```
set base  https://azuresearch-usnc.nuget.org/query
```

Which resulted in the error `Must specify a server`

An extra space before the url should not impact the command. Other dotnet commands ignore extra spaces between parameters.

image

@bradygaster @tlmii

tlmii:
Fixes #365 

This is a short-term quick fix to try to sneak it into 3.1 while we work on the new parsing in 5.0. Will cherry-pick to release/3.1 once approved.

### DIFF
--- a/src/Microsoft.Repl/Parsing/CoreParser.cs
+++ b/src/Microsoft.Repl/Parsing/CoreParser.cs
@@ -14,6 +14,9 @@ namespace Microsoft.Repl.Parsing
             commandText = commandText ?? throw new ArgumentNullException(nameof(commandText));
 
             List<string> sections = commandText.Split(' ').ToList();
+            // We can't use StringSplitOptions.RemoveEmptyEntries because it
+            // is more aggressive than we need, so we need to do it ourselves.
+            RemoveEmptyEntries(sections);
             Dictionary<int, int> sectionStartLookup = new Dictionary<int, int>();
             HashSet<int> quotedSections = new HashSet<int>();
             int runningIndex = 0;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Repl/Parsing/CoreParser.cs
```diff
@@ -14,6 +14,9 @@ public ICoreParseResult Parse(string commandText, int caretPosition)
             commandText = commandText ?? throw new ArgumentNullException(nameof(commandText));
 
             List<string> sections = commandText.Split(' ').ToList();
+            // We can't use StringSplitOptions.RemoveEmptyEntries because it
+            // is more aggressive than we need, so we need to do it ourselves.
+            RemoveEmptyEntries(sections);
```
tlmii, line 19: Using `StringSplitOptions.RemoveEmptyEntries` will remove empty entries at the end of the string and we have a good bit of code and tests that rely on empty entries at the end and provide specific error messages for those scenarios. To keep those intact, I rolled my own.

src/Microsoft.Repl/Parsing/CoreParser.cs
```diff
@@ -131,5 +134,23 @@ public ICoreParseResult Parse(string commandText, int caretPosition)
...
+
+        private static void RemoveEmptyEntries(List<string> sections)
+        {
+            if (sections.Count < 2)
+            {
+                return;
+            }
+
+            // We want to remove empty spaces from the beginning, and from the middle
+            // but not from the end.
```
bzhu94, line 146: Any reason to not remove it from the end?

tlmii: See my comment on the PR - https://github.com/dotnet/HttpRepl/pull/366#discussion_r463159983 - there's a lot of assumptions around it, including about how we handle suggestions (if there is no space, tab would cycle suggestions on the last keyword; if there was a space, it would be trying to handle the next keyword or option/arg).

sayedihashimi:
@tlmii thanks for fixing this, it will be a great improvment

PR #376, opened 2020-08-10
Replace custom OpenAPI parsing with OpenAPI.NET

### LINKED ISSUE #347: Use OpenAPI.NET?
### Change Details

We currently have three separate custom parsers for Swagger/OpenAPI - `SwaggerV1ApiDefinitionReader`, `SwaggerV2ApiDefinitionReader` and `OpenApiV3ApiDefinitionReader`. We can remove those three parsers and make use of the Reader functionality in `Microsoft.OpenApi.Readers`. 

> **Note**: This means we lose support for Swagger V1, but it was determined through discussions below hat was acceptable.

One added benefit, beyond the removal of a lot of code, is that we get support for YAML descriptions "for free", something we'd have needed to add additional parsers for anyway.

### Original Issue:

There's an existing parser library - [OpenAPI.NET](https://github.com/microsoft/openapi.net) - that we could make use of instead of our custom parser. I'm not positive why it wasn't used in the first place, but best guess is just timing - I believe that when the original parser code was written for HttpRepl, OpenAPI.NET was still under initial development.

From an initial glance at the project, the only downside I see is that it doesn't currently support the v1 spec, and HttpRepl does. But that could be fixed by either a) contributing the v1 implementation to that project, thus helping everyone, or b) dropping v1 support from HttpRepl, since usage is likely low (v2 is almost 6 years old at this point).

On the plus side, it would mean removing the majority of the code in [/src/Microsoft.HttpRepl/OpenApi/](https://github.com/dotnet/HttpRepl/tree/master/src/Microsoft.HttpRepl/OpenApi), would likely save us from having to fix some bugs in the parser (I'm expecting that #345 is one of those, for example) and anything that we _do_ run into, we could push the fix upstream and help everyone.

@timheuer The v1 question probably goes through you first. Also cc @darrelmiller in case he has any data or info that could help us on that decision.

tlmii:
TL;DR: Closes #347 and replaces our custom Swagger/OpenAPI parser implementations with [OpenAPI.NET](https://github.com/microsoft/OpenAPI.NET).

See #347 for details of the implications here.


Directory.Build.props
```diff
@@ -1,9 +1,13 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project>
   <!-- Leave this file here, even if it's empty. It stops chaining imports. -->
 
   <Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk" />
 
+  <PropertyGroup>
+    <LangVersion>preview</LangVersion>
+  </PropertyGroup>
```
tlmii, line 9: Went ahead and moved us to C# 9, mostly so I could use `is not null`. It'll be out of preview by the time we release 5.0.0, so we'll be able to remove this.

eng/Signing.props
```diff
@@ -6,5 +6,6 @@
   <ItemGroup>
     <FileSignInfo Include="Newtonsoft.Json.dll" CertificateName="3PartySHA2" />
     <FileSignInfo Include="Newtonsoft.Json.Bson.dll" CertificateName="3PartySHA2" />
+    <FileSignInfo Include="SharpYaml.dll" CertificateName="3PartySHA2" />
```
tlmii, line 9: This is pulled in via `Microsoft.OpenApi.Readers`

global.json
```diff
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "5.0.100-preview.6.20310.4"
+    "dotnet": "5.0.100-preview.7.20366.6"
```
tlmii, line 3: Needed to bump this for the C# 9 code to work on the build machines.

src/Microsoft.HttpRepl/OpenApi/OpenApiDotNetApiDefinitionReader.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            if (openApiDocument.Paths is not null)
+            {
+                foreach (KeyValuePair<string, OpenApiPathItem> path in openApiDocument.Paths)
+                {
+                    string relativeUrl = path.Key;
+
+                    EndpointMetadata endpointMetadata = new EndpointMetadata(relativeUrl);
+
+                    foreach (KeyValuePair<OperationType, OpenApiOperation> operation in path.Value.Operations)
```
jimmylewis, line 66: nit: Thoughts on pulling some of these foreach loops out into separate functions for readability?  So we'd end up with something like:

```
// this block is now nice and short
if (openApiDocument.Paths is not null)
{
    foreach (KeyValuePair<string, OpenApiPathItem> path in openApiDocument.Paths)
    {
        string relativeUrl = path.Key;
        EndpointMetadata endpointMetadata = new EndpointMetadata(relativeUrl);
        BuildEndpointOperations(endpointMetadata, path.Value);
        metadata.Add(endpointMetadata);
    }
}

// this goes somewhere else now
internal void BuildEndpointOperations(EndpointMetadata endpointMetadata, OpenApiPathItem path)
{
    foreach (KeyValuePair<OperationType, OpenApiOperation> operation in path.Value.Operations)
    {
        RequestMetadata requestMetadata = new RequestMetadata(operation.Key);

        foreach (OpenApiParameter parameter in operation.Value.Parameters)
        {
            requestMetadata.Parameters.Add(parameter);
        }

        if (operation.Value.RequestBody?.Content is not null)
        {
            foreach (KeyValuePair<string, OpenApiMediaType> content in operation.Value.RequestBody.Content)
            {
                string contentType = content.Key;
                bool isRequired = operation.Value.RequestBody.Required;
                OpenApiSchema schema = content.Value?.Schema;

                requestMetadata.Content.Add(new RequestContentMetadata(contentType, isRequired, schema));
            }
        }

        endpointMetadata.AvailableRequests.Add(requestMetadata);
    }
}
```

src/Microsoft.HttpRepl/OpenApi/OpenApiDotNetApiDefinitionReader.cs
```diff
@@ -0,0 +1,108 @@
...
+                        }
+
+                        endpointMetadata.AvailableRequests.Add(requestMetadata);
+                    }
+
+                    metadata.Add(endpointMetadata);
+                }
+            }
+
+            DirectoryStructure d = new DirectoryStructure(null);
```
jimmylewis, line 94: This could also be pulled out into a separate utility method, e..g `apiDefinition.DirectoryStructure = BuildDirectoryStructure(metadata);`

test/Microsoft.HttpRepl.Tests/Microsoft.HttpRepl.Tests.csproj
```diff
@@ -1,10 +1,32 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>net5.0</TargetFramework>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
+  <ItemGroup>
```
jimmylewis, line 8: I can't recall if it's necessary to do the remove or whether the SDK got smarter about this, but you could try to reduce duplication here by doing something like:
```
<ItemGroup>
    <EmbeddedResource Include="..." /> <!-- repeated for many files -->
    <None Remove="@(EmbeddedResource)" /> <!-- just remove any embedded resource from the None group -->
</ItemGroup>
```

src/Microsoft.HttpRepl/OpenApi/Either.cs
```diff
@@ -1,46 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-namespace Microsoft.HttpRepl.OpenApi
-{
-    public class Either<TOption1, TOption2>
```
bzhu94, line 6: This class seems really wacky...Glad you're removing it :)

src/Microsoft.HttpRepl/OpenApi/SchemaDataGenerator.cs
```diff
@@ -3,50 +3,51 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
+using Microsoft.OpenApi.Models;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.HttpRepl.OpenApi
 {
     public static class SchemaDataGenerator
```
bzhu94, line 11: Is it OK to leave this as public?

src/Microsoft.HttpRepl/OpenApi/SchemaDataGenerator.cs
```diff
@@ -3,50 +3,51 @@
...
     {
-        public static string GetBodyString(IEnumerable<Parameter> operation)
-        {
-            Parameter body = operation.FirstOrDefault(x => string.Equals(x.Location, "body", StringComparison.OrdinalIgnoreCase));
+        // TODO: Make this a setting?
+        private const int MaxExampleDataDepth = 3;
```
bzhu94, line 14: Just curious, what problems did we face to have this included in here? Is this an arbitrarily picked number?

src/Microsoft.HttpRepl/OpenApi/ApiDefinitionReader.cs
```diff
@@ -3,25 +3,22 @@
 
 using System;
 using System.Collections.Generic;
-using Newtonsoft.Json.Linq;
 
 namespace Microsoft.HttpRepl.OpenApi
 {
-    public class ApiDefinitionReader
+    internal class ApiDefinitionReader
```
bzhu94, line 9: So we expect no consumers of Http Repl to depend on this class (and other public classes in this repo?)

src/Microsoft.HttpRepl/OpenApi/OpenApiDotNetApiDefinitionReader.cs
```diff
@@ -0,0 +1,108 @@
...
+using Microsoft.OpenApi.Models;
+using Microsoft.OpenApi.Readers;
+
+namespace Microsoft.HttpRepl.OpenApi
+{
+    internal class OpenApiDotNetApiDefinitionReader : IApiDefinitionReader
+    {
+        public bool CanHandle(string document)
+        {
+            OpenApiStringReader reader = new();
```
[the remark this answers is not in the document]
tlmii: Yep. Amusingly I wrote most of this code (or one of them, not sure which) with `var` and then had VS convert it to explicit... and then remembered C# 9 had target-typed new. So that one line of code went through a lot of change for being so simple.

src/Microsoft.HttpRepl/OpenApi/SchemaDataGenerator.cs
```diff
@@ -73,7 +74,7 @@ internal static JToken GenerateData(Schema schema)
                             return (schema.Maximum.Value + schema.Minimum.Value) / 2;
                         }
 
-                        if (schema.ExclusiveMinimum)
+                        if (schema.ExclusiveMinimum == true)
```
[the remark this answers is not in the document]
tlmii: It seems to.

src/Microsoft.HttpRepl/OpenApi/ApiDefinitionReader.cs
```diff
@@ -3,25 +3,22 @@
 
 using System;
 using System.Collections.Generic;
-using Newtonsoft.Json.Linq;
 
 namespace Microsoft.HttpRepl.OpenApi
 {
-    public class ApiDefinitionReader
```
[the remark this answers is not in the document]
tlmii: Since the HttpRepl part (as opposed to the Repl part) is a command line executable, not a library, we don't really expect there to be consumers of it, in a programmatic sense. If we wanted to make it reusable, we'd probably split that off into another layer and make the appropriate things public there.

src/Microsoft.HttpRepl/OpenApi/ApiDefinitionReader.cs
tlmii: See comment above

src/Microsoft.HttpRepl/OpenApi/EndpointMetadata.cs
```diff
@@ -2,19 +2,51 @@
...
 using System.Collections.Generic;
+using Microsoft.OpenApi.Models;
 
 namespace Microsoft.HttpRepl.OpenApi
 {
-    public class EndpointMetadata
+    internal class EndpointMetadata
     {
-        public EndpointMetadata(string path, IReadOnlyDictionary<string, IReadOnlyDictionary<string, IReadOnlyList<Parameter>>> requestsByMethodAndContentType)
+        public EndpointMetadata(string path)//, IReadOnlyDictionary<string, IReadOnlyDictionary<string, IReadOnlyList<Parameter>>> requestsByMethodAndContentType)
```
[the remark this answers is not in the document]
tlmii: Whoops. Got it.

src/Microsoft.HttpRepl/OpenApi/OpenApiDotNetApiDefinitionReader.cs
tlmii: I went one step further and split out the Path stuff first, then the Operations stuff from that. Had been following the structure of the old V3 ApiDefinitionrReader and never went back to clean it up. Probably need to do a little more for efficiency, but this is probably good enough for now readability wise at least.

test/Microsoft.HttpRepl.Tests/Microsoft.HttpRepl.Tests.csproj
tlmii: VS added the removes as I switched each file to Embedded resource, but they don't seem to be necessary as far as I can tell. I removed them and everything seems to work and not complain.

src/Microsoft.HttpRepl/OpenApi/SchemaDataGenerator.cs
tlmii: I switched it to internal like the others I changed. There's probably more that could be internal, but I'll make another pass at that later.

src/Microsoft.HttpRepl/OpenApi/SchemaDataGenerator.cs
tlmii: Arbitrary-ish. Did some testing looking at what the JSON would look like if we went any deeper than that and it got a little difficult to follow if we got past 3, especially if there were many paths that went past 3.

The primary problem that brought it up was actually a stack overflow exception caused by cycles. In one test scenario, it just kept trying to generate a loop of objects that eventually referred back to a parent object type. You don't necessarily need a loop for that to happen, though, if your object hierarchies can really be that complicated. But rather than handle the cycles by tracking and breaking if they happen AND handling deep hierarchies with a limit, I just went with a limit.

The reality is, hardly anyone (I learned never to say no one) is going to hand-edit something more than a couple objects deep. They might have something pre-generated that they paste into the editor over top of this, but then we've just wasted cycles generating it. 

I still think it should be user configurable eventually so that if someone really wanted it for their scenario they could get a few levels deeper. But that's for another PR.

bradygaster:
this is exciting to see. making sure @darrelmiller see this fancy new update

darrelmiller:
Woot!  This is awesome.

PR #386, opened 2020-09-10
Fixup /swagger root urls

tlmii:
As of .NET 5 RC1, the `dotnet new webapi` template now has a default start url of `swagger`. Because the default Swashbuckle-generated OpenAPI description doesn't contain a Servers element, this will put HttpRepl users that launch from VS/VS Code into a pit of failure by having a base address of `https://localhost:{port}/swagger/`, even though the API is, by default, based at the root. The user won't be able to navigate their API at all, because you can't navigate "above" the base address.

Since it is unlikely a user would put their API inside the `/swagger` path, we will special-case this scenario and remove that from the url. We will give the user an escape hatch via the preference if they do put their API under that path.


bradygaster (approved):
Excellent and cc @coolcsh fyi

bzhu94 (approved):
Looks good!

PR #387, opened 2020-09-11
Add Telemetry

tlmii:
This PR adds telemetry to the tool. The underlying implementation is based heavily on the telemetry implementation in [dotnet-try](https://github.com/dotnet/try/tree/main/Microsoft.DotNet.Interactive.Telemetry). That includes the informational banner (run once each time the tool is updated) and the ability to opt-out of telemetry via an environment variable.

Every event collects some basic information (same as dotnet-try):
- The tool version
- The OS platform and version
- The kernel version
- The runtime id
- Whether or not it is being run in a docker container
- A machine id, generated and hashed in the same way as Visual Studio

We track the following events:
- When the tool starts, we record whether the user tried to use the `run` command, the `help` command, the `connect` command or just opened the tool with no args. We do not collect the arguments themselves.
- When a command is executed, we record which command was executed and whether or not the command succeeded
- When the `connect` command is executed, we record whether the root url, base url and/or openapi url were specified and whether or not we could locate the openapi description. We do not collect the actual urls used.
- When the `pref` command is executed, we record whether it was a `get` or a `set` and which preference was being read/modified. If the preference specified is not a well-known preference, we hash the preference name. We do not collect the values being set.
- When the `set header` command is executed, we record the name of the header being set. If the header is in the well-known headers list, we record it verbatim. If it is not, the value is hashed. We do not collect the values of the headers being set
- When the `connect` command is executed with a root url that ends in `"/swagger"`, we record whether it was modified or if the modification was skipped (per #386). We do not record the actual url.


src/Microsoft.HttpRepl.Telemetry/UserLevelCacheWriter.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        public string RunWithCache(string cacheKey, Func<string> getValueToCache)
+        {
+            var cacheFilepath = GetCacheFilePath(cacheKey);
+            try
+            {
+                if (!_fileExists(cacheFilepath))
+                {
+                    if (!_directoryExists(_dotnetTryUserProfileFolderPath))
```
jimmylewis, line 56: nit: Without abstracting away the file system, calling `Directory.CreateDirectory` is cheap and safe and doesn't require checking if it exists first.

src/Microsoft.HttpRepl.Telemetry/UserLevelCacheWriter.cs
```diff
@@ -0,0 +1,89 @@
...
+                    return getValueToCache();
+                }
+
+                throw;
+            }
+        }
+
+        private string GetCacheFilePath(string cacheKey)
+        {
+            return Path.Combine(_dotnetTryUserProfileFolderPath, $"{_productVersion}_{cacheKey}.dotnetTryUserLevelCache");
```
jimmylewis, line 86: `dotnetTryUserLevelCache` should be `dotnetHttpReplUserLevelCache`?

src/Microsoft.HttpRepl.Telemetry/UserLevelCacheWriter.cs
```diff
@@ -0,0 +1,89 @@
...
+                Directory.Exists,
+                path => Directory.CreateDirectory(path),
+                File.WriteAllText,
+                File.ReadAllText)
+        {
+        }
+
+        public UserLevelCacheWriter(
+            string productVersion,
+            string dotnetTryUserProfileFolderPath,
```
jimmylewis, line 33: Try should be HttpRepl?

src/Microsoft.HttpRepl.Telemetry/ITelemetry.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.HttpRepl.Telemetry
+{
+    public interface ITelemetry
+    {
+        bool Enabled { get; }
```
jimmylewis, line 10: nit: `IsEnabled` - I think we usually prefer the `Is` prefix on state properties.

src/Microsoft.HttpRepl/Commands/ClearCommand.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.HttpRepl.Commands
 {
     public class ClearCommand : ICommand<object, ICoreParseResult>
     {
-        private static readonly string Name = "clear";
+        public string Name => "clear";
```
jimmylewis, line 16: nit: arrange this with the other public members instead of with the private field?

src/Microsoft.HttpRepl/Commands/SetHeaderCommand.cs
```diff
@@ -18,14 +20,22 @@ namespace Microsoft.HttpRepl.Commands
 {
     public class SetHeaderCommand : ICommand<HttpState, ICoreParseResult>
     {
-        private static readonly string Name = "set";
+        private static readonly string CommandName = "set";
         private static readonly string SubCommand = "header";
 
+        private readonly ITelemetry _telemetry;
+
+        public string Name => "SetHeader";
```
jimmylewis, line 28: Is this the only command using PascalCase for the Name property?  all the others have been lowercase.

src/Microsoft.HttpRepl/Program.cs
```diff
@@ -22,18 +23,25 @@ public class Program
     {
         public static async Task Main(string[] args)
         {
-            await new Program().Start(args);
+            await Start(args);
         }
 
-        public async Task Start(string[] args, IConsoleManager consoleManager = null, IPreferences preferences = null)
+        internal static async Task Start(string[] args, IConsoleManager consoleManager = null, IPreferences preferences = null, ITelemetry telemetry = null)
```
jimmylewis, line 29: Why the change to internal?  This was left public to force the IntegrationTests to not have an IVT, so that they wouldn't dabble with internal types.

src/Microsoft.HttpRepl/Telemetry/TelemetryExtensions.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.HttpRepl.Telemetry.Events;
+
+namespace Microsoft.HttpRepl.Telemetry
+{
+    internal static class TelemetryExtensions
+    {
+        public static void TrackEvent(this ITelemetry telemetry, TelemetryEventBase telemetryEvent)
```
jimmylewis, line 10: I was wondering why this isn't just a method on ITelemetry, is it because it's defined in the Telemetry assembly but TelemetryEventBase is defined in the HttpRepl assembly?  (If so, I'm fine with this, just wanted to clarify the need to implement via extension method)

src/Microsoft.Repl/Commanding/ICommand.cs
```diff
@@ -11,6 +11,8 @@ namespace Microsoft.Repl.Commanding
     public interface ICommand<in TProgramState, in TParseResult>
         where TParseResult : ICoreParseResult
     {
+        string Name { get; }
```
jimmylewis, line 14: nit: maybe add an XMLDoc comment explaining what Name is used for?

jimmylewis (approved):
Orthogonal to this change, I noticed there's a lot of string handling involving nulls.  Have we thought about enabling nullable checks in this project?

tlmii:
> Orthogonal to this change, I noticed there's a lot of string handling involving nulls. Have we thought about enabling nullable checks in this project?

@jimmylewis Yep, I've already got a branch where I'm working on that. This PR was just higher priority, but that change is coming soon.

src/Microsoft.HttpRepl/Telemetry/TelemetryExtensions.cs
tlmii: Yep, that's right. Keeping the HttpRepl-specific stuff (like TelemetryEventBase) separate from the shared-by-source pieces that are in the Telemetry assembly.

src/Microsoft.HttpRepl.Telemetry/MacAddressGetter.cs
```diff
@@ -0,0 +1,191 @@
...
+            foreach (Match match in Regex.Matches(shelloutput, MacRegex, RegexOptions.IgnoreCase))
+            {
+                if (!Regex.IsMatch(match.Value, ZeroRegex))
+                {
+                    macAddress = match.Value;
+                    break;
+                }
+            }
+
+            if (macAddress != null)
```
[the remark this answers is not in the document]
tlmii: I went ahead and streamlined this since it definitely helped readability but didn't change functionality.

src/Microsoft.HttpRepl.Telemetry/MacAddressGetter.cs
```diff
@@ -0,0 +1,191 @@
...
+                    if (ifconfigResult == 0)
+                    {
+                        return ifconfigStdOut;
+                    }
+                    else
+                    {
+                        return GetIpCommandOutput();
+                    }
+                }
+                catch (Win32Exception e)
```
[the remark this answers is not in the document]
tlmii: I would hope not... but looking back through the history in dotnet/try and dotnet/sdk, its been that way since the beginning, so I'm hesitant to change it right now.

src/Microsoft.HttpRepl.Telemetry/MacAddressGetter.cs
```diff
@@ -0,0 +1,191 @@
...
+                    return null;
+                }
+            }
+            else
+            {
+                try
+                {
+                    var ifconfigResult = new ProcessStartInfo
+                    {
+                        FileName = "ifconfig",
```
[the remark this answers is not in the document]
tlmii: My understanding is that the order that all of this is done in is done to mimic the way existing tools do it to make sure the hashed machine ids can line up. So I don't think we want to change that (yet).

src/Microsoft.HttpRepl.Telemetry/Telemetry.cs
```diff
@@ -0,0 +1,202 @@
...
+            }
+        }
+
+        public bool Enabled { get; }
+
+        public IFirstTimeUseNoticeSentinel FirstTimeUseNoticeSentinel { get; }
+
+        public static bool SkipFirstTimeExperience => GetEnvironmentVariableAsBool(HttpRepl.Telemetry.FirstTimeUseNoticeSentinel.SkipFirstTimeExperienceEnvironmentVariableName, false);
+
+        public static bool IsRunningInDockerContainer => GetEnvironmentVariableAsBool("DOTNET_RUNNING_IN_CONTAINER", false);
```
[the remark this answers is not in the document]
tlmii: Agreed, especially since it was only called from `DockerContainerDetectorForTelemetry` anyway. I moved it.

src/Microsoft.HttpRepl.Telemetry/Telemetry.cs
```diff
@@ -0,0 +1,202 @@
...
+
+        private static string PrependProducerNamespace(string eventName)
+        {
+            return "dotnet/httprepl/" + eventName;
+        }
+
+        private Dictionary<string, double> GetEventMeasures(IReadOnlyDictionary<string, double> measurements)
+        {
+            Dictionary<string, double> eventMeasurements = new Dictionary<string, double>(_commonMeasurements);
+            if (measurements != null)
```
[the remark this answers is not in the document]
tlmii: I don't know. Interestingly, both were implemented the same way originally, then only GetEventProperties was changed and not this one - and the commit says it was in response to PR feedback. I can't seem to find the PR though, so I'm hesitant to change it even though I don't why it would matter.

src/Microsoft.HttpRepl.Telemetry/TelemetryCommonProperties.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        public TelemetryCommonProperties(
+            string productVersion,
+            Func<string, string> hasher = null,
+            Func<string> getMACAddress = null,
+            IDockerContainerDetector dockerContainerDetector = null,
+            IUserLevelCacheWriter userLevelCacheWriter = null)
+        {
+            _productVersion = productVersion;
+            _hasher = hasher ?? Sha256Hasher.Hash;
```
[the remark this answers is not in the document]
tlmii: For all the non-telemetry pieces (plus the ITelemetry/Telemetry object itself) we do have the ComposeDependencies in Program.cs that organizes them normally. I didn't think it was necessary to pull all of these up into that, especially given the hope that this would eventually be consolidated into a separate shared assembly which may or may not have that same pattern.

src/Microsoft.HttpRepl.Telemetry/TelemetryCommonProperties.cs
```diff
@@ -0,0 +1,121 @@
...
+                return _dockerContainerDetector.IsDockerContainer().ToString("G");
+            });
+        }
+
+        /// <summary>
+        /// Returns a string identifying the OS kernel.
+        /// For Unix this currently comes from "uname -srv".
+        /// For Windows this currently comes from RtlGetVersion().
+        ///
+        /// Here are some example values:
```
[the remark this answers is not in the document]
tlmii: Removed.

src/Microsoft.HttpRepl/Commands/SetHeaderCommand.cs
tlmii: There was one other - I changed them both to camel for consistency

src/Microsoft.HttpRepl/Program.cs
tlmii: It was because code analysis complained about not checking `consoleManager`, `preferences` and `telemetry` for `null` before using them, since they were in the public interface. Setting aside that they _were_ checked before use, in `ComposeDependencies`, it wasn't _really_ a public interface in my mind, so I changed it to `internal`. But I didn't really like having the IVT either. So I changed it back and added the suppression attribute instead. Hopefully I can get that analysis warning to go away once I add nullable annotations (though my quick try didn't work - code analysis didn't seem to care about the `[NotNull]` attributes on `ComposeDependencies` parameters).

src/Microsoft.HttpRepl.Telemetry/Telemetry.cs
```diff
@@ -0,0 +1,202 @@
...
+                _client.Context.Device.OperatingSystem = RuntimeEnvironment.OperatingSystem;
+
+                _commonProperties = new TelemetryCommonProperties(productVersion).GetTelemetryCommonProperties();
+                _commonMeasurements = new Dictionary<string, double>();
+            }
+            catch (Exception e)
+            {
+                _client = null;
+                // we dont want to fail the tool if telemetry fails.
+                Debug.Fail(e.ToString());
```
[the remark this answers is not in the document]
tlmii: I added #388 because I think that is something that could be useful even outside of this narrow scenario. But we don't have any logging to disk right now at all, so I didn't want to try to cram that in here.

tlmii:
@jimmylewis If I didn't respond to one of your comments above about something in the Telemetry assembly, it was because of our discussion offline -wanting to keep the code fairly similar to the original sources. I did make some changes if it was obviously not a functional change or didn't deviate too far from the structure.

bzhu94 (approved):
Looks good!

Issue #372, opened 2020-08-05
Command to Validate OpenAPI Description?


tlmii:
With the work we're doing to move to using OpenApi.NET (#347), one of the benefits we get is the ability to validate the OpenAPI Description document. 

I'm wondering if it might be useful to have a command that validates the document and writes out the errors, if any. Something like:

```
Disconnected> openapi validate https://localhost/api/openapi.json
Validating OpenAPI Description at https://localhost/api/openapi.json
Error: "XYZ is not a valid property of Path"
Error: "The info object is required."
```

We could also make the validation be a part of connecting, if we enable a more verbose mode for connect.

Most people will be using Swashbuckle or something like that to generate the description, so they should have a valid document. But if they're not, and they're seeing weird results within httprepl, this would be a helpful way to diagnose the problems.

bradygaster:
love this idea. i think it'd be best if it were a separate command that someone could run **and** as a configurable switch [defaulted to false] that someone could turn on. That way they could pre-validate any calls, or just use it to validate.

tlmii:
> We could also make the validation be a part of connecting, if we enable more verbose mode for connect

This part is done, so users have the ability to get validation results. We could still decide to make it a separate command. I could see this being used as part of a testing/validation script - several consecutive `openapi validate [url]` calls to ensure there are no errors.

sayedihashimi:
**Thank you for your interest in HttpRepl.**

This project has been deprecated, and the repository is now archived. As part of this transition, we are closing all open issues.
Please read the https://github.com/dotnet/HttpRepl/issues/701 for alternative recommendations.

PR #405, opened 2020-10-21
Add --verbosity option to connect command

### LINKED ISSUE #373: Verbosity setting for connect command?
Right now, the `connect` command is pretty quiet in terms of the output. It tells you the results (which base address and open api address it picked) when it finishes, but otherwise there is no real output.

I propose that we have add a verbose output mode for the command. So the two output modes would be:

- Standard: Same as now plus adding validation information (see #372) for the selected OpenAPI description if one was found.
- Verbose: Log all of the OpenAPI description endpoints we try and why they were not selected (404 Not Found or 401 Not Authorized status codes, or no ApiDefinitionReader could handle what was returned) along with validation errors.

Standard would be the default and we could add a `--verbose|-v` option to the command.

@bradygaster This is more or less what I was talking about when we discussed the SSL issues you were having. Right now the tool just fails silently. This would at least point the user in the right direction even if we can't fix it for them.

tlmii:
Resolves #373.

Adds a `--verbose` (or `-v`) option to the `connect` command that (initially) gives more details about the process of finding the OpenAPI Description.

To do this, I implemented a wrapper around the `IConsoleManager` that has a flag for whether verbosity is enabled. This kept us from having `if (isVerbose) { writeToLog(...) }` kinds of checks all over the code. This is a first cut of that and we'll probably expand it a bit as we add more logging and verbosity options across the tool.

We'll add more information to the verbose output in the future around OpenAPI validation (related to #372 likely). But this was the focus for the initial push.


bradygaster (approved):

src/Microsoft.HttpRepl/VerbosityLogger.cs
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#nullable enable
+
+using Microsoft.Repl.ConsoleHandling;
+
+namespace Microsoft.HttpRepl
+{
+    internal class VerbosityLogger
```
jimmylewis, line 10: Having a concrete class named VerbosityLogger referenced throughout the code is feeling awkward to me.  To any consuming code, it should be just a Logger, or better yet, an ILogger.  The implementation can be a WhateverLogger, but that should be resolved at composition time (or maybe an ILoggerFactory for more dynamic acquisition cases).

Also, the acquisition of the logger, which is currently a static method, means that you have hidden dependencies strewn in the consuming code.  It's not following good DI principles.

src/Microsoft.HttpRepl/ApiConnection.cs
```diff
@@ -81,31 +80,51 @@ public async Task FindSwaggerDoc(HttpClient client, IEnumerable<string> swaggerS
...
+                else
+                {
+                    logger.WriteLineVerbose(response.StatusCode.ToString());
+                    return (false, null);
+                }
             }
-            catch
+            catch (Exception e)
             {
+                logger.WriteLineVerbose(e.Message);
```
jimmylewis, line 121: What's the error experience when not using verbose?  Should we always show all error details in non-verbose?

src/Microsoft.HttpRepl/ApiConnection.cs
```diff
@@ -118,19 +137,19 @@ public void SetupApiDefinition(HttpState programState)
             }
         }
 
-        public async Task SetupHttpState(HttpState httpState, bool performAutoDetect, CancellationToken cancellationToken)
+        public async Task SetupHttpState(HttpState httpState, bool performAutoDetect, VerbosityLogger logger, CancellationToken cancellationToken)
```
jimmylewis, line 140: Would you ever pass different loggers to a single instance of this?  I wonder if it would be nicer to take as a construction parameter, and then you don't have to pass this through every call.

src/Microsoft.HttpRepl/VerbosityLogger.cs
```diff
@@ -0,0 +1,72 @@
...
+        private readonly bool _isVerbosityEnabled;
+        private readonly IConsoleManager? _consoleManager;
+
+        private VerbosityLogger(IConsoleManager? consoleManager, bool isVerbosityEnabled)
+        {
+            _consoleManager = consoleManager;
+            _isVerbosityEnabled = isVerbosityEnabled;
+        }
+
+        public static VerbosityLogger FromConsoleManager(IConsoleManager? consoleManager, bool isVerbosityEnabled = false)
```
jimmylewis, line 21: What does a static factory method get you that a public constructor wouldn't, when they take the same args?  It's a more confusing API to have to discover that I can't just new() it up.

test/Microsoft.HttpRepl.Tests/Commands/GetCommandTests.cs
```diff
@@ -139,12 +140,18 @@ public async Task ExecuteAsync_WithJsonContentTypeInHeader_FormatsResponseConten
     }
   }
 }";
+
+            string[] expectedResponseLines = expectedResponse.Split(Environment.NewLine);
+
             List<string> result = shellState.Output;
 
-            Assert.Equal(3, result.Count);
+            Assert.Equal(2 + expectedResponseLines.Length, result.Count);
```
jimmylewis, line 148: the 2 here is magic sauce from non-Logger output?  Can we comment that or put it into a well-named constant (since the magic 2 is used on line 153 as well)?

test/Microsoft.HttpRepl.Tests/VerbosityLoggerTests.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using Microsoft.HttpRepl.Fakes;
+using Xunit;
+
+namespace Microsoft.HttpRepl.Tests
+{
+    public class VerbosityLoggerTests
+    {
+        [Fact]
+        public void Writes_WithNullConsoleManager_DoesNotThrow()
```
jimmylewis, line 12: Why is a null ConsoleManager allowed anyway?

src/Microsoft.HttpRepl/ApiConnection.cs
tlmii: There's two scenarios that use this code. The first, when you specify an OpenAPI Description directly. The second, when we're using the list of possible OpenAPI Description paths to check a bunch of them to find the first one that works. 

Currently (prior to this PR) the end result is the same in both scenarios - either it found a swagger file, or it didn't and we say where we found it if we did. WHY we didn't find it (or didn't find several) is not shown. 

For the former case, I could see an argument for always showing the why. For the latter, we wouldn't want the user to see a bunch of failures (for however many times it fails before it succeeds) - that's where the verbose use case came from. 

So, at least currently, its a little odd because the exception isn't necessarily unexpected depending on the codepath. Maybe I can enable verbose for the directly-specified scenario automatically, with the need for the option?

test/Microsoft.HttpRepl.Tests/Commands/GetCommandTests.cs
tlmii: None of it is from logger output. This was because of how I changed MockShellState to record to a StringBuffer instead of a List<string>. The json was originally, incorrectly, being written as "one" line in that List<string>, even though it had line breaks in it. So when I changed it to a StringBuffer with splitting, that made it have the correct number of actual output lines. The 2 is for the two lines checked directly below this line (`HTTP/1.1 200 OK` and `Content-Type: application/json`). I can pull it out to a variable anyway since its used twice.

src/Microsoft.HttpRepl/VerbosityLogger.cs
tlmii: This basically stems from a really poor attempt by me at pulling this particular feature out of a larger effort to enable a sort of "Verbosity Context" for commands that would allow a command to turn on verbosity (with an option or otherwise) and have it flow to any child commands or components. VerbosityLogger was kind of a stripped down version of the work in progress there, but I did a poor job shoehorning it in. 

I've removed that, and simplified the implementation here (at least, I think). I still don't like the tight coupling of specifying the verbosity as it goes into the ApiConnection, but its not awful and its better than specifying the VerbosityLogger in those places.

src/Microsoft.HttpRepl/ApiConnection.cs
tlmii: I went ahead and enabled verbose for ApiConnection when specifying the swagger file directly, which I think is helpful and in the spirit of the ask for the related issue anyway.

test/Microsoft.HttpRepl.Tests/Commands/GetCommandTests.cs
```diff
@@ -139,12 +141,19 @@ public async Task ExecuteAsync_WithJsonContentTypeInHeader_FormatsResponseConten
     }
   }
 }";
+
+            string[] expectedResponseLines = expectedResponse.Split(Environment.NewLine);
+            
+
```
jimmylewis, line 147: nit: two empty lines

src/Microsoft.HttpRepl/ApiConnection.cs
```diff
@@ -183,5 +188,29 @@ private IEnumerable<string> GetSwaggerSearchPaths()
                 return paths;
             }
         }
+
+        private void WriteVerbose(string s)
```
jimmylewis, line 192: I still don't like this.  Whether it's verbose or not should be settled between the command handler and the logger, and should not be a responsibility of this type; `ApiConnection` should just use LogMessage and LogVerboseMessage equivalents. 

To put it another way, `ApiConnection` should know what is normal vs. verbose output, but needn't have any logic or switching behavior about verbosity.  Under my proposed design, the command handler (e.g. BaseHttpCommand, ConnectCommand, HelpCommand, ListCommand, etc) should configure the logger (or IWriteable now) as to whether to show verbose output, then invoke ApiCommand.  

Two hypothetical examples (snippet borrowed from ConnectCommand):
```
bool logVerboseMessages = isVerbosityEnabled || !string.IsNullOrWhiteSpace(swaggerAddress);
using (shellState.ConsoleManager.ShowVerboseMessages(logVerboseMessages))
{
    ApiConnection apiConnection = new ApiConnection(preferences, shellState.ConsoleManager);
    if (!string.IsNullOrWhiteSpace(rootAddress))	
    { ... }
}

// or...
bool logVerboseMessages = isVerbosityEnabled || !string.IsNullOrWhiteSpace(swaggerAddress);
var loggerWithVerbosity = shellState.ConsoleManager.GetWriteable(verbosity: logVerboseMessages);
ApiConnection apiConnection = new ApiConnection(preferences, loggerWithVerbosity);
```

You've mentioned that this area needs a deeper look, so I'm fine with punting this to a later issue/PR to refactor it around.  But let's have a conversation at some point to see what this might look like.

jimmylewis (approved):

jimmylewis (approved):

src/Microsoft.HttpRepl/ApiConnection.cs
tlmii: Your first example with the using statement is fairly close in concept to what I had been prototyping for the overall system. The only real big difference is that it needs to be in another layer. ConsoleManager is just one possible output location for the "logger", whatever it may be called. File output is another (that will be more important with the upcoming scripting focus) and, potentially, "variables" is a third (mostly for output of http commands, but not necessarily limited to that). Its not clear yet how much of that will be (or even should be) the same logger, or needs for certain to have the same verbosity requirements or anything like that. But the console manager just being one output location seems correct.

So that was where the VerbosityLogger came in originally - it was sort of that intermediate component, but massively stripped down to near uselessness for this purpose... including not having a good creation story because the rest of the plumbing wasn't there to have it available. Thus the awkward creation in a couple places. That was a bad idea on my part, likely just because I had so much of that other design in my head.

So on this second cut, I just went barebones. ApiConnection knows about the switching right now, but in as few places as possible. In the end, we'll probably be changing the constructor to drop the boolean and take an IWhateverLogger instead of an IWritable and change the calls to .Write/.WriteLine to be on that object instead.

Dfinitely want to have a design discussion around this once I get more clarification on some of the things that are in the air for after v5 - scripting and output to files (and anything else that goes along with making it usable for more than just ad hoc testing) are likely high on the priority list and fit into that discussion.

src/Microsoft.HttpRepl/ApiConnection.cs
```diff
@@ -81,31 +85,51 @@ public async Task FindSwaggerDoc(HttpClient client, IEnumerable<string> swaggerS
             }
         }
 
-        public async Task<string> GetSwaggerDocAsync(HttpClient client, Uri uri, CancellationToken cancellationToken)
+        private async Task<(bool Success, string Document)> GetSwaggerDocAsync(HttpClient client, Uri uri, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
tlmii: Good excuse to just `#nullable enable` this file and get it out of the way!

PR #411, opened 2020-10-26
Move request details from help to verbose list command

### LINKED ISSUE #70: Split HelpCommand output
Currently, the HelpCommand has three main code paths when executed:

1. If you just say `help`, it displays a canned and curated lists of commands and their help summaries
2. If you say `help <command> [<subCommand>]`, where <command> is whatever command you're interested in and <subCommand> is an optional subCommand for the command, then you'll get the detailed help for that command (assuming it exists).
3. If you say `help <relativeUrl>`, where <relativeUrl> is a path to a known api endpoint, it displays a more verbose version of what you would have gotten if you (starting at the root) did `cd <relativeUrl>` followed by `ls`. Note that this relativeUrl is always relative to the root, not to the current directory.

For this issue I'm proposing splitting off that third option since it seems to be in the wrong place, and instead making an optional subcommand or argument on the `ls` command, something like `ls verbose` that displays the information in the same format as the current help implementation does, but following the semantics of the current `ls` command where it looks at the current directory.

Not saying this needs to be done now, Just wanted to put an issue out there to track it.

cc @jimmylewis @bhsubra @jodavis for your thoughts.

### LINKED ISSUE #409: Commands with more than one command name form do not parse correctly
Commands with more than one command name form (e.g. the List command, which can either be `ls` or `dir`) are not parsed correctly. The first n command segments are skipped (where n is the number of forms in the command name spec minus one - so usually this means skipping just one segment).

### LINKED ISSUE #410: Recursive form of the list command mixes the case of request methods in the output
The first level (which is there with or without recursion) is cased correctly (all uppercase). But everything else is title cased.

tlmii:
Resolves #70 by moving the form of the help command that gave details about a particular route out of `help` and into a verbose variant of the list command (`ls --verbose` or `dir --verbose`).

Along the way, also resolves #409 since that prevented this from working as well and resolves #410 since that became immediately apparent once #409 was fixed.

### DIFF
--- a/src/Microsoft.HttpRepl/Commands/HelpCommand.cs
+++ b/src/Microsoft.HttpRepl/Commands/HelpCommand.cs
@@ -7,7 +7,6 @@ using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.HttpRepl.Preferences;
 using Microsoft.HttpRepl.Resources;
 using Microsoft.HttpRepl.Suggestions;
 using Microsoft.HttpRepl.Telemetry;
@@ -22,11 +21,9 @@ namespace Microsoft.HttpRepl.Commands
     {
         public string Name => "help";
 
-        private readonly IPreferences _preferences;
-
-        public HelpCommand(IPreferences preferences)
+        public HelpCommand()
         {
-            _preferences = preferences;
+            
         }
 
         public bool? CanHandle(IShellState shellState, HttpState programState, ICoreParseResult parseResult)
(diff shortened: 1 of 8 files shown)


src/Microsoft.HttpRepl/Commands/HelpCommand.cs
```diff
@@ -94,77 +94,14 @@ public async Task ExecuteAsync(IShellState shellState, HttpState programState, I
...
-                                    }
-                                }
-                            }
-                        }
-
-                        if (!anyHelp)
-                        {
-                            output.AppendLine("Unable to locate any help information for the specified command");
-                        }
+                        output.AppendLine("Unable to locate any help information for the specified command");
```
jimmylewis, line 97: Should this string be localized?

jimmylewis (approved):

PR #412, opened 2020-10-26
Warn about command arguments when using VS Code

### LINKED ISSUE #310: Does not wait for editor in some cases when doing a POST/PUT/etc
HttpRepl version: 3.0.47301
VS Code version: 1.40.2
OS: Windows 10

If I have VS Code open, and VS Code is set to be the editor on the tool, the "template" file pops open in VS Code correctly, but the tab title for the file says "(deleted)" and the tool does not wait for the file to be closed, it just sends it right away and then deletes the file. 

If I don't have VS Code open, and so it is launched when executing a POST, the process seems to work as expected.

tlmii:
Resolves #310 

Since, at this point, we don't want to override the user's settings automatically, for now we will just present a warning if we think they are setting their default editor to VS Code. The warning guides them to set the proper command arguments.

### DIFF
--- a/src/Microsoft.HttpRepl/Commands/PrefCommand.cs
+++ b/src/Microsoft.HttpRepl/Commands/PrefCommand.cs
@@ -276,15 +276,18 @@ namespace Microsoft.HttpRepl.Commands
         {
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
-                return path.Contains("Code.exe", StringComparison.OrdinalIgnoreCase);
+                return path.Contains("Code.exe", StringComparison.OrdinalIgnoreCase) ||
+                       path.Contains("Code - Insiders.exe", StringComparison.OrdinalIgnoreCase);
             }
             else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
             {
-                return path.Contains("Visual Studio Code.app", StringComparison.Ordinal);
+                return path.Contains("Visual Studio Code.app", StringComparison.Ordinal) ||
+                       path.Contains("Visual Studio Code - Insiders.app", StringComparison.Ordinal);
             }
             else //  Linux
             {
-                return string.Equals(path, "/usr/bin/code", StringComparison.Ordinal);
+                return string.Equals(path, "/usr/bin/code", StringComparison.Ordinal) ||
+                       string.Equals(path, "/usr/bin/code-insiders", StringComparison.Ordinal);
             }
         }
     }
--- a/src/Microsoft.HttpRepl/Commands/PrefCommand.cs
+++ b/src/Microsoft.HttpRepl/Commands/PrefCommand.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -177,6 +178,19 @@ namespace Microsoft.HttpRepl.Commands
             {
                 shellState.ConsoleManager.Error.WriteLine(Resources.Strings.PrefCommand_Error_Saving.SetColor(programState.ErrorColor));
             }
+            else
+            {
+                // If we think they're configurating HttpRepl to use Visual Studio Code as their editor, we should
+                // warn them that for best integration, they should also pass the `-w` or `--wait` arguments to
+                // Visual Studio Code.
+                if (string.Equals(prefName, WellKnownPreference.DefaultEditorCommand, StringComparison.Ordinal))
+                {
+                    if (IsVSCode(prefValue))
+                    {
+                        shellState.ConsoleManager.WriteLine(string.Format(Resources.Strings.PrefCommand_Set_VSCode, WellKnownPreference.DefaultEditorArguments).SetColor(programState.WarningColor));
+                    }
+                }
+            }
         }
 
         private void GetSetting(IShellState shellState, HttpState programState, DefaultCommandInput<ICoreParseResult> commandInput)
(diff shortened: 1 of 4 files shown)


src/Microsoft.HttpRepl/Commands/PrefCommand.cs
```diff
@@ -177,6 +178,19 @@ private void SetSetting(IShellState shellState, HttpState programState, DefaultC
             {
                 shellState.ConsoleManager.Error.WriteLine(Resources.Strings.PrefCommand_Error_Saving.SetColor(programState.ErrorColor));
             }
+            else
+            {
+                // If we think they're configurating HttpRepl to use Visual Studio Code as their editor, we should
+                // warn them that for best integration, they should also pass the `-w` or `--wait` arguments to
```
bradygaster, line 184: 👍

src/Microsoft.HttpRepl/Commands/PrefCommand.cs
```diff
@@ -257,5 +271,21 @@ protected override IEnumerable<string> GetArgumentSuggestionsForText(IShellState
...
+
+        private static bool IsVSCode(string path)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                return path.Contains("Code.exe", StringComparison.OrdinalIgnoreCase);
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                return path.Contains("Visual Studio Code.app", StringComparison.Ordinal);
```
bradygaster, line 283: Do we care if they're using insiders?

bradygaster (commented):
My only concern would be if we make accommodations for VS Code, but opt out of doing so for VS Code Insiders, if some folks will ask for that anyway. Myself, I **am** the customer for whom i'm advocating in this review, as I *only* use VS Code Insiders now and don't even have the regular version installed.

src/Microsoft.HttpRepl/Commands/PrefCommand.cs
```diff
@@ -177,6 +178,19 @@ private void SetSetting(IShellState shellState, HttpState programState, DefaultC
...
                 shellState.ConsoleManager.Error.WriteLine(Resources.Strings.PrefCommand_Error_Saving.SetColor(programState.ErrorColor));
             }
+            else
+            {
+                // If we think they're configurating HttpRepl to use Visual Studio Code as their editor, we should
+                // warn them that for best integration, they should also pass the `-w` or `--wait` arguments to
+                // Visual Studio Code.
+                if (string.Equals(prefName, WellKnownPreference.DefaultEditorCommand, StringComparison.Ordinal))
+                {
+                    if (IsVSCode(prefValue))
```
jimmylewis, line 188: Should/can we also check if they have already set the argument correctly?  E.g. if I'm switching from Code to Code-Insiders, maybe I don't need the warning.  Or if I set the arguments before the editor.  Showing the warning will make me double-check it, and possibly waste my time.  (This is a pretty nitty case though)

src/Microsoft.HttpRepl/Commands/PrefCommand.cs
```diff
@@ -257,5 +271,24 @@ protected override IEnumerable<string> GetArgumentSuggestionsForText(IShellState
...
+                       path.Contains("Code - Insiders.exe", StringComparison.OrdinalIgnoreCase);
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                return path.Contains("Visual Studio Code.app", StringComparison.Ordinal) ||
+                       path.Contains("Visual Studio Code - Insiders.app", StringComparison.Ordinal);
+            }
+            else //  Linux
+            {
+                return string.Equals(path, "/usr/bin/code", StringComparison.Ordinal) ||
```
jimmylewis, line 289: Doing a full path equals seems like we could miss cases.  What if it's in `~/bin` or `/usr/local/bin` or some variant?  (I don't know if anyone would do this, but since we only check executable name on the other two platforms, this one stands out...)

tlmii:
@bradygaster Agree on insiders. I've updated this to include the insiders paths for all three operating systems. I'm guessing this detection will change over time (people can, theoretically, install it to other places). But this should be a good start and we'll get it into the documentation as well.

jimmylewis (approved):

src/Microsoft.HttpRepl/Commands/PrefCommand.cs
tlmii: I went back and forth on this. I was worried about just doing, essentially, `.Contains("code")` thinking it might catch false positives. The others OSs have more explicit things we can check for. I guess maybe it could be `.EndsWith("code")` and that might have fewer false positives? I think we're already talking about a small number of cases... just not sure which is the right slightly smaller number of cases :)

jimmylewis: Or an `.EndsWith("/bin/code")` might catch a few more, without pinning solely to the /usr mount point.  But it is trying to catch a small handful of edge cases, so feel free to punt on it too. 😉

tlmii: I expect that this will evolve over time as we get feedback from users, so I think we can start here and see what happens. It might need to change even further if we ever change the tool so the whole path isn't required (so the command could just be "code" on basically all platforms and it'd work). Right now that doesn't work because of how we use `ProcessStartInfo`/`Process.Start` (with `ShellExecute` false being the issue I believe) but I _think_ there are some options there. Just hasn't been a priority yet.

src/Microsoft.HttpRepl/Commands/PrefCommand.cs
tlmii: For the moment, I'm going to say that inconvenience is small enough that it doesn't warrant trying to parse someone's existing arguments (especially since `-w` could be inside another argument (e.g. `--new-window`) or a quoted argument (no e.g. off the top of my head, but we'd want to avoid it if we were actually parsing)).

PR #417, opened 2020-10-27
Update Analyzers to 3.3 and disable var

### LINKED ISSUE #413: Update Analyzers to 3.0
Maybe include more analyzers as well?

tlmii:
Resolves #413, fixes a bunch of warnings/suggestions (I mostly skipped the test and telemetry projects for now) and also switches the .editorconfig var setting to suggest the explicit type name rather than suggesting var.

Added #418 to track one error that I wasn't positive about how to fix at the moment, but didn't want to block the rest of this.

### DIFF
--- a/src/Microsoft.HttpRepl.Telemetry/UserLevelCacheWriter.cs
+++ b/src/Microsoft.HttpRepl.Telemetry/UserLevelCacheWriter.cs
@@ -48,7 +48,7 @@ namespace Microsoft.HttpRepl.Telemetry
 
         public string RunWithCache(string cacheKey, Func<string> getValueToCache)
         {
-            getValueToCache = getValueToCache ?? throw new ArgumentNullException(nameof(getValueToCache));
+            _ = getValueToCache ?? throw new ArgumentNullException(nameof(getValueToCache));
 
             var cacheFilepath = GetCacheFilePath(cacheKey);
             try
--- a/src/Microsoft.HttpRepl.Telemetry/UserLevelCacheWriter.cs
+++ b/src/Microsoft.HttpRepl.Telemetry/UserLevelCacheWriter.cs
@@ -48,6 +48,8 @@ namespace Microsoft.HttpRepl.Telemetry
 
         public string RunWithCache(string cacheKey, Func<string> getValueToCache)
         {
+            getValueToCache = getValueToCache ?? throw new ArgumentNullException(nameof(getValueToCache));
+
             var cacheFilepath = GetCacheFilePath(cacheKey);
             try
             {
--- a/src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
+++ b/src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
@@ -158,9 +158,9 @@ namespace Microsoft.HttpRepl.Commands
                     }
                 }
 
-                var responseFileOption = commandInput.Options[ResponseFileOption].Any() ? commandInput.Options[ResponseFileOption][0] : null;
-                var responseHeadersFileOption = commandInput.Options[ResponseHeadersFileOption].Any() ? commandInput.Options[ResponseHeadersFileOption][0] : null;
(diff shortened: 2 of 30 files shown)


src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
```diff
@@ -457,7 +463,9 @@ private static async Task FormatBodyAsync(DefaultCommandInput<ICoreParseResult>
                     {
                         try
                         {
+#pragma warning disable CA2012 // Use ValueTasks correctly
```
jimmylewis, line 466: Oof, why is this complaining?  ValueTasks are fragile creatures if not used correctly.  You must consume their result once, and there are possible perf implications (or bugs) if you don't.  See https://github.com/dotnet/roslyn-analyzers/issues/3617#issuecomment-625986124 and https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/#valid-consumption-patterns-for-valuetasks.

src/Microsoft.HttpRepl.Telemetry/UserLevelCacheWriter.cs
```diff
@@ -48,6 +48,8 @@ public sealed class UserLevelCacheWriter : IUserLevelCacheWriter
 
         public string RunWithCache(string cacheKey, Func<string> getValueToCache)
         {
+            getValueToCache = getValueToCache ?? throw new ArgumentNullException(nameof(getValueToCache));
```
jimmylewis, line 51: fun fact: `_ = getValueToCache ?? throw new ArgumentNullException(nameof(getValueToCache));` will save you 2 ops in MSIL, IIRC.

tlmii: I like it.

src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
tlmii: Yeah, it seemed to be complaining for a valid reason. I'm just not sure yet exactly how I can back out its usage. `WaitForCompletionAsync` checks `readTask` to see if it is completed in a loop (which seems bad, based on the documentation) while also checking to see if there are any keystrokes available in the console and bailing when either thing happens. 
I _think_ I can just drop that altogether and let the cancellation token be the only thing that cancels the read (the token should get canceled if you ctrl+c during the operation). But I wanted to do some testing on that specifically and didn't think that should block the rest of the PR. That's why I added #418 to make sure I came back to it.

tlmii: Actually, looking at the second link with the scenarios you shouldn't do, I'm not sure why it's complaining. I thought it was the multiple checks on IsCompleted and the multiple checks on Result. But those aren't called out. It does seem to fail the short rule though:
> The short rule is this: with a ValueTask or a ValueTask<TResult>, you should either await it directly (optionally with .ConfigureAwait(false)) or call AsTask() on it directly, and then never use it again

Regardless, I think I'll be able to remove that extra work and just await it. Just need to confirm some things separately.

jimmylewis: I'd be willing to bet that the problem is that you're passing it off to another method - it is a struct after all, so you're not consuming the ValueTask in this scope, you're consuming the copy-by-value'd clone of it in the other method.  (At least, you don't consume it in all code paths here... the `.Result` call is inside an if).

If you're going to pass it around, you should probably `.AsTask()` it first, and then it would behave as expected.

(Really, the fact that you're passing a ValueTask around at all seems like a bug that violates the implicit ValueTask contract.)

tlmii: I went ahead and removed the suppression and used `.AsTask()` to make it safer while I figure out if I can remove the extra waits and checks for keystrokes. Updated #418 a little bit to note what I'm doing.

jimmylewis (approved):

PR #419, opened 2020-10-27
Handle HTTP Request errors in BaseHttpCommand

### LINKED ISSUE #304: httprepl nice idea, but why are error messages like fail to connect shown as stacktraces?
Issue originally opened by @mattiasw2 at https://github.com/aspnet/AspNetCore.Docs/issues/15936.

> For example, instead of "No connection ...." I get this
> 
> http://localhost:5000/&gt; options
> System.Net.Http.HttpRequestException: No connection could be made because the target machine actively refused it.
>  ---&gt; System.Net.Sockets.SocketException (10061): No connection could be made because the target machine actively refused it.
>    at System.Net.Http.ConnectHelper.ConnectAsync(String host, Int32 port, CancellationToken cancellationToken)
>    --- End of inner exception stack trace ---
>    at System.Net.Http.ConnectHelper.ConnectAsync(String host, Int32 port, CancellationToken cancellationToken)
>    at System.Net.Http.HttpConnectionPool.ConnectAsync(HttpRequestMessage request, Boolean allowHttp2, CancellationToken cancellationToken)
> :
> :

tlmii:
Resolves #304.

Handle some of the "expected" exceptions in `BaseHttpCommand` when executing HTTP Requests via `HttpClient`.

### DIFF
--- a/src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
+++ b/src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
@@ -165,8 +165,21 @@ namespace Microsoft.HttpRepl.Commands
                 string headersTarget = responseHeadersFileOption?.Text ?? responseFileOption?.Text;
                 string bodyTarget = responseBodyFileOption?.Text ?? responseFileOption?.Text;
 
-                HttpResponseMessage response = await programState.Client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
-                await HandleResponseAsync(programState, commandInput, shellState.ConsoleManager, response, programState.EchoRequest, headersTarget, bodyTarget, cancellationToken).ConfigureAwait(false);
+                try
+                {
+                    HttpResponseMessage response = await programState.Client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
+                    await HandleResponseAsync(programState, commandInput, shellState.ConsoleManager, response, programState.EchoRequest, headersTarget, bodyTarget, cancellationToken).ConfigureAwait(false);
+                }
+                catch (HttpRequestException httpRequestException)
+                {
+                    shellState.ConsoleManager.Error.WriteLine(httpRequestException.Message.SetColor(programState.ErrorColor));
+                }
+                catch (OperationCanceledException)
+                {
+                    // We just want to eat this exception because the cancellation actually occurs for the entire command,
+                    // not just the HTTP Request. So the cancellation is handled further down the stack by inspecting
+                    // the CancellationToken.IsCancellationRequested property
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
```diff
@@ -165,8 +165,21 @@ protected override async Task ExecuteAsync(IShellState shellState, HttpState pro
                 string headersTarget = responseHeadersFileOption?.Text ?? responseFileOption?.Text;
                 string bodyTarget = responseBodyFileOption?.Text ?? responseFileOption?.Text;
 
-                HttpResponseMessage response = await programState.Client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
-                await HandleResponseAsync(programState, commandInput, shellState.ConsoleManager, response, programState.EchoRequest, headersTarget, bodyTarget, cancellationToken).ConfigureAwait(false);
```
bzhu94, line 169: Just curious, if nothing was catching exceptions here, what thing did catch them to allow the exception stack to show in the console?

src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
```diff
@@ -165,8 +165,21 @@ protected override async Task ExecuteAsync(IShellState shellState, HttpState pro
...
+                    await HandleResponseAsync(programState, commandInput, shellState.ConsoleManager, response, programState.EchoRequest, headersTarget, bodyTarget, cancellationToken).ConfigureAwait(false);
+                }
+                catch (HttpRequestException httpRequestException)
+                {
+                    shellState.ConsoleManager.Error.WriteLine(httpRequestException.Message.SetColor(programState.ErrorColor));
+                }
+                catch (OperationCanceledException)
+                {
+                    // We just want to eat this exception because the cancellation actually occurs for the entire command,
+                    // not just the HTTP Request. So the cancellation is handled further down the stack by inspecting
```
bzhu94, line 180: So before this fix the user would see a stacktrace anytime they canceled a command? That's not good... 😝

bzhu94 (approved):
Looks good!

src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
tlmii: Further up the command execution pipeline:
https://github.com/dotnet/HttpRepl/blob/feb5c12bc9c2d6d72667826d1a5fb2f36ab73a87/src/Microsoft.Repl/Commanding/DefaultCommandDispatcher.cs#L119-L122

src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
tlmii: > So before this fix the user would see a stacktrace anytime they canceled a command?

Depends on the command. But if there was an HTTP request in flight, yes. 
> that's not good... 😝

Nope!

PR #429, opened 2020-11-05
Remove --response option from HTTP commands

### LINKED ISSUE #426: Remove --response option from HTTP Commands
Per #407 it doesn't actually work, and has never worked. The user will still have `--response:Body` and `--response:Header` that do work. But we don't want the help and documentation pointing to something they can't actually use.

The only real functional change here will be ensuring that the user doesn't specify the same file for both - that was the long-form equivalent of `--response`, is also broken and, per #404, we don't really have a good idea of what the structure of that file should be anyway. So we can just error out if they do try that.

tlmii:
Resolves #426.

Also fixed a bug in the parser where it didn't handle quoted sections that didn't have spaces in them. The fix is maybe not a complete fix, but I didn't think it was worth going further since we have the `System.CommandLine` work coming up that will hopefully replace the entire parser.

### DIFF
--- a/src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
+++ b/src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
@@ -672,7 +672,9 @@ namespace Microsoft.HttpRepl.Commands
 
         protected override IEnumerable<string> GetOptionValueCompletions(IShellState shellState, HttpState programState, string optionId, DefaultCommandInput<ICoreParseResult> commandInput, ICoreParseResult parseResult, string normalizedCompletionText)
         {
-            if (string.Equals(optionId, BodyFileOption, StringComparison.Ordinal) || string.Equals(optionId, ResponseBodyFileOption, StringComparison.OrdinalIgnoreCase) || string.Equals(optionId, ResponseHeadersFileOption, StringComparison.OrdinalIgnoreCase))
+            if (string.Equals(optionId, BodyFileOption, StringComparison.Ordinal) ||
+                string.Equals(optionId, ResponseBodyFileOption, StringComparison.OrdinalIgnoreCase) ||
+                string.Equals(optionId, ResponseHeadersFileOption, StringComparison.OrdinalIgnoreCase))
             {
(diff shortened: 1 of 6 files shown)


src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
```diff
@@ -677,7 +672,7 @@ protected override IEnumerable<string> GetArgumentSuggestionsForText(IShellState
 
         protected override IEnumerable<string> GetOptionValueCompletions(IShellState shellState, HttpState programState, string optionId, DefaultCommandInput<ICoreParseResult> commandInput, ICoreParseResult parseResult, string normalizedCompletionText)
         {
-            if (string.Equals(optionId, BodyFileOption, StringComparison.Ordinal) || string.Equals(optionId, ResponseFileOption, StringComparison.OrdinalIgnoreCase) || string.Equals(optionId, ResponseBodyFileOption, StringComparison.OrdinalIgnoreCase) || string.Equals(optionId, ResponseHeadersFileOption, StringComparison.OrdinalIgnoreCase))
+            if (string.Equals(optionId, BodyFileOption, StringComparison.Ordinal) || string.Equals(optionId, ResponseBodyFileOption, StringComparison.OrdinalIgnoreCase) || string.Equals(optionId, ResponseHeadersFileOption, StringComparison.OrdinalIgnoreCase))
```
jimmylewis, line 675: nit: for the sake of readability, maybe make this multiple lines?
```
if (string.Equals(optionId, BodyFileOption, StringComparison.Ordinal) || string.Equals(optionId, ResponseBodyFileOption, StringComparison.OrdinalIgnoreCase) || string.Equals(optionId, ResponseHeadersFileOption, StringComparison.OrdinalIgnoreCase))
//vs
if (string.Equals(optionId, BodyFileOption, StringComparison.Ordinal) 
    || string.Equals(optionId, ResponseBodyFileOption, StringComparison.OrdinalIgnoreCase) 
    || string.Equals(optionId, ResponseHeadersFileOption, StringComparison.OrdinalIgnoreCase))
```

jimmylewis (approved):

PR #430, opened 2020-11-05
Add telemetry for HTTP commands

### LINKED ISSUE #427: Add telemetry for arguments used with HTTP Commands
Right now we know that the user is issuing GET, POST, etc. But we don't know if they're taking advantage of any of the other arguments (e.g. `--response:Body`, `--response:Header`).

For the eventual discussions we'll have around priorities for the post-5.0 release, this will be valuable to have.

tlmii:
Resolves #427 

Adds a new telemetry event sent by all of the HTTP commands collecting which options were specified with the command. The values of any options are not collected.


jimmylewis (changes requested):

jimmylewis (approved):
Per discussion, we'll revisit this after the System.CommandLine work

PR #438, opened 2020-11-11
Move telemetry doc content to ASP.NET Core docs

scottaddie:
The telemetry Markdown file's content was published to docs yesterday. See https://github.com/dotnet/AspNetCore.Docs/pull/20468. We want the doc to be the source of truth, since that content is localized to numerous languages. Updating this Markdown file to point to the doc.

/cc: @tlmii


TELEMETRY.md
```diff
@@ -1,48 +1,3 @@
-# HttpRepl telemetry
+# HTTP REPL telemetry
```
tlmii, line 1: We need to get a ruling on this from someone (@bradygaster ?) - the tool itself mostly refers to itself as HttpRepl, but the docs mostly say HTTP REPL. For this one change, it doesn't matter. But we should make a pass for consistency once we find a way to standardize it.

tlmii (approved):

TELEMETRY.md
bradygaster: HttpRepl is my vote

Issue #258, opened 2019-10-11
Support for query parameters


Leon99:
Usability of this tool would be so much higher if it supported listing and specifying query parameters.

tlmii:
I agree - definitely something I want to look into. Thanks for flagging it.

salihkarabilo:
Could you prioritize this ? As I think the most common case to use Httprepl is when running a simple get request with some query string values.

bravecobra:
With query parameter support, this tool is unusable for us. We use FromQuery all the time in our api's

bradygaster:
@tlmii this feels like something that we'll need to prioritize in the next iteration. What say you?

tlmii:
@bradygaster I'm definitely fine with prioritizing it. We just have lots of questions to answer before we can put more code down (other than #448).

We have the information from the OpenAPI Description to say what the available query string parameters are for a given endpoint. So that's helpful. But what we need to figure out is how does the user specify them? 
- Is there a corollary to the `cd` way of drilling in through the path parameters that we could use with query string parameters? 
- Is it just an option on the various HTTP commands (effectively the per-call equivalent of the global command in #448)?
- Do we display it as part of the prompt?
- Should we display them as part of the results of an `ls`?

I'd be curious to know from the others in this thread (@bravecobra @salihkarabilo @Leon99 ) what their use cases are from the three scenarios I outlined in https://github.com/dotnet/HttpRepl/pull/448#issuecomment-735010282. An interesting data point to glean from folks is just how often they're using per-call query string parameters vs global.

bravecobra:
To clarify my particular use-case here: 
I want to use the HttpRepl for scripting out user-scenarios for integration testing, a series of api calls to be executed so to say. Something like, 

- add a product
- get the product
- update the product description
- search for a partial description and verify the product is found
- update the product price
- get the product again and verify the price has been updated.

Since our api's use query parameters, that was a no-go up until now. Currently I'm looking into an efficient way to do this, either use the HttpRepl and script it out, or use PostMan or setup a C# unittest with a generated C# client from the openapi spec. I have to rule out the HttpRepl for now.

Note that query parameters are not always required, listing them out as part of the `ls` command would require listing out every possible combination when multiple query parameters are involved. I'd imagine it would be handy to list out all the available ones when describing one particular route, but not on a list of routes. In other words, I don't use the httprepl as an  api-browser, but I want to be able to make any call the api exposes. Hope this helps.

[...]

PR #448, opened 2020-11-25
Add QueryString command support

ygalbel:
I created queryString command support.

The code is very similar to setHeader command.
I support multiple time the same key (`like ?key=value1&key=value2`).

The value is always url encoded.
All the code is tested.

I didn't implement suggestions.

### DIFF
--- a/src/Microsoft.HttpRepl/Commands/HelpCommand.cs
+++ b/src/Microsoft.HttpRepl/Commands/HelpCommand.cs
@@ -186,6 +186,7 @@ namespace Microsoft.HttpRepl.Commands
             output.AppendLine();
             output.AppendLine($"{"connect",navCommandColumn}{GetCommand<ConnectCommand>(dispatcher).GetHelpSummary(shellState, programState)}");
             output.AppendLine($"{"set header",navCommandColumn}{GetCommand<SetHeaderCommand>(dispatcher).GetHelpSummary(shellState, programState)}");
+            output.AppendLine($"{"set queryString",navCommandColumn}{GetCommand<SetQueryStringCommand>(dispatcher).GetHelpSummary(shellState, programState)}");
 
             output.AppendLine();
             output.AppendLine(Strings.HelpCommand_Core_HttpCommands.Bold().Cyan());
--- a/src/Microsoft.HttpRepl/Commands/SetQueryStringCommand.cs
+++ b/src/Microsoft.HttpRepl/Commands/SetQueryStringCommand.cs
@@ -1,0 +1,91 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.HttpRepl.Resources;
+using Microsoft.HttpRepl.Suggestions;
+using Microsoft.HttpRepl.Telemetry;
+using Microsoft.HttpRepl.Telemetry.Events;
+using Microsoft.Repl;
+using Microsoft.Repl.Commanding;
+using Microsoft.Repl.ConsoleHandling;
+using Microsoft.Repl.Parsing;
+
+namespace Microsoft.HttpRepl.Commands
+{
+    public class SetQueryStringCommand : ICommand<HttpState, ICoreParseResult>
+    {
+        private const string CommandName = "set";
+        private const string SubCommand = "queryString";
+
+        private readonly ITelemetry _telemetry;
+
+        public string Name => "setQueryString";
+        public static string Description => Strings.SetQueryStringCommand_HelpDetails;
+
+        public SetQueryStringCommand(ITelemetry telemetry)
+        {
+            _telemetry = telemetry;
+        }
+
+        public bool? CanHandle(IShellState shellState, HttpState programState, ICoreParseResult parseResult)
+        {
+            return parseResult.ContainsAtLeast(minimumLength: 3, CommandName, SubCommand)
+                ? (bool?)true
+                : null;
+        }
+
+        public Task ExecuteAsync(IShellState shellState, HttpState programState, ICoreParseResult parseResult, CancellationToken cancellationToken)
+        {
+            parseResult = parseResult ?? throw new ArgumentNullException(nameof(parseResult));
+
+            programState = programState ?? throw new ArgumentNullException(nameof(programState));
+
+            bool isValueEmpty;
+            if (parseResult.Sections.Count == 3)
+            {
+                programState.QueryString.Remove(parseResult.Sections[2]);
+                isValueEmpty = true;
+            }
+            else
+            {
+                programState.QueryString[parseResult.Sections[2]] = parseResult.Sections.Skip(3);
+                isValueEmpty = false;
+            }
+
+            _telemetry.TrackEvent(new SetQueryStringEvent(parseResult.Sections[2], isValueEmpty));
+
+            return Task.CompletedTask;
+        }
+
+        public string GetHelpDetails(IShellState shellState, HttpState programState, ICoreParseResult parseResult)
+        {
+            if (parseResult.ContainsAtLeast(CommandName, SubCommand))
+            {
+                StringBuilder helpText = new StringBuilder();
+                helpText.Append(Strings.Usage.Bold());
+                helpText.AppendLine("set queryString {name} [value]");
+                helpText.AppendLine();
+                helpText.AppendLine(Strings.SetQueryStringCommand_HelpDetails);
+                return helpText.ToString();
+            }
+
+            return null;
+        }
+
+        public string GetHelpSummary(IShellState shellState, HttpState programState)
+        {
+            return Description;
+        }
+
+        public IEnumerable<string> Suggest(IShellState shellState, HttpState programState, ICoreParseResult parseResult)
+        {
+            return null;
+        }
+    }
+}
--- a/src/Microsoft.HttpRepl/HttpState.cs
+++ b/src/Microsoft.HttpRepl/HttpState.cs
@@ -91,13 +91,12 @@ namespace Microsoft.HttpRepl
             return GetEffectivePath(BaseAddress, string.Join('/', PathSections.Reverse()), commandSpecifiedPath, null);
         }
 
-
-        public Uri GetEffectivePathWithQueryString(string commandSpecifiedPath, Dictionary<string,IEnumerable<string>> queryString)
+        public Uri GetEffectivePathWithQueryString(string commandSpecifiedPath, Dictionary<string, IEnumerable<string>> queryString)
         {
             return GetEffectivePath(BaseAddress, string.Join('/', PathSections.Reverse()), commandSpecifiedPath, queryString);
         }
 
-        public static Uri GetEffectivePath(Uri baseAddress, string pathSections, string commandSpecifiedPath, Dictionary<string,IEnumerable<string>> queryString)
+        public static Uri GetEffectivePath(Uri baseAddress, string pathSections, string commandSpecifiedPath, Dictionary<string, IEnumerable<string>> queryString)
         {
             // If an absolute uri string was already specified, just return that.
             if (Uri.IsWellFormedUriString(commandSpecifiedPath, UriKind.Absolute))
@@ -121,7 +120,6 @@ namespace Microsoft.HttpRepl
 
             if (queryString != null)
             {
-                
                 foreach (KeyValuePair<string, IEnumerable<string>> tuple in queryString)
                 {
                     foreach (var singleValue in tuple.Value)
--- a/src/Microsoft.HttpRepl/HttpState.cs
+++ b/src/Microsoft.HttpRepl/HttpState.cs
@@ -35,7 +35,7 @@ namespace Microsoft.HttpRepl
 
         public Dictionary<string, IEnumerable<string>> Headers { get; }
 
-        public Dictionary<string, IEnumerable<string>> QueryString{ get; }
+        public Dictionary<string, IEnumerable<string>> QueryString { get; }
 
         public Uri SwaggerEndpoint { get; set; }
 
--- a/src/Microsoft.HttpRepl/HttpState.cs
+++ b/src/Microsoft.HttpRepl/HttpState.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
 using System.Net.Http;
 using Microsoft.HttpRepl.FileSystem;
 using Microsoft.HttpRepl.OpenApi;
@@ -34,6 +35,8 @@ namespace Microsoft.HttpRepl
 
         public Dictionary<string, IEnumerable<string>> Headers { get; }
 
+        public Dictionary<string, IEnumerable<string>> QueryString{ get; }
+
         public Uri SwaggerEndpoint { get; set; }
 
         public HttpState(IPreferences preferences, HttpClient httpClient)
@@ -47,6 +50,7 @@ namespace Microsoft.HttpRepl
             {
                 { "User-Agent", new[] { _preferences.GetValue(WellKnownPreference.HttpClientUserAgent, "HTTP-REPL") } }
             };
+            QueryString = new Dictionary<string, IEnumerable<string>>();
         }
 
         public string GetPrompt()
@@ -84,10 +88,16 @@ namespace Microsoft.HttpRepl
 
         public Uri GetEffectivePath(string commandSpecifiedPath)
         {
-            return GetEffectivePath(BaseAddress, string.Join('/', PathSections.Reverse()), commandSpecifiedPath);
+            return GetEffectivePath(BaseAddress, string.Join('/', PathSections.Reverse()), commandSpecifiedPath, null);
         }
 
-        public static Uri GetEffectivePath(Uri baseAddress, string pathSections, string commandSpecifiedPath)
+
+        public Uri GetEffectivePathWithQueryString(string commandSpecifiedPath, Dictionary<string,IEnumerable<string>> queryString)
+        {
+            return GetEffectivePath(BaseAddress, string.Join('/', PathSections.Reverse()), commandSpecifiedPath, queryString);
+        }
+
+        public static Uri GetEffectivePath(Uri baseAddress, string pathSections, string commandSpecifiedPath, Dictionary<string,IEnumerable<string>> queryString)
         {
             // If an absolute uri string was already specified, just return that.
             if (Uri.IsWellFormedUriString(commandSpecifiedPath, UriKind.Absolute))
@@ -99,6 +109,7 @@ namespace Microsoft.HttpRepl
             {
                 throw new ArgumentNullException(nameof(baseAddress), string.Format(Resources.Strings.HttpState_Error_NoAbsoluteUriNoBaseAddress, nameof(commandSpecifiedPath), nameof(baseAddress)));
             }
+            
 
             pathSections = pathSections ?? throw new ArgumentNullException(nameof(pathSections));
 
@@ -107,6 +118,18 @@ namespace Microsoft.HttpRepl
             UpdateUriBuilderForSpecifiedPath(builder, commandSpecifiedPath, out string commandQuery);
             AppendQueryToBuilder(builder, baseAndPathQuery);
             AppendQueryToBuilder(builder, commandQuery);
+
+            if (queryString != null)
+            {
+                
+                foreach (KeyValuePair<string, IEnumerable<string>> tuple in queryString)
+                {
+                    foreach (var singleValue in tuple.Value)
+                    {
+                        AppendQueryToBuilder(builder, $"{tuple.Key}={WebUtility.UrlEncode(singleValue)}");
+                    }
+                }
+            }
 
             return builder.Uri;
         }
@@ -118,7 +141,7 @@ namespace Microsoft.HttpRepl
                 return null;
             }
 
-            return GetEffectivePath(BaseAddress, string.Join('/', PathSections.Reverse()), "");
+            return GetEffectivePath(BaseAddress, string.Join('/', PathSections.Reverse()), "", null);
         }
 
         public string GetRelativePathString()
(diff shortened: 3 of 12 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/HttpRepl?pullRequest=448) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/HttpRepl?pullRequest=448) before we can accept your contribution.<br/><br/>:x: ygalbel [sign now](https://cla.dotnetfoundation.org/dotnet/HttpRepl?pullRequest=448)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/HttpRepl?pullRequest=448) it.</sub>

tlmii:
@ygalbel Thanks for taking the time to submit the PR!

I'll want input from @bradygaster, @sayedihashimi, etc when we get past the long holiday weekend. 

But in the interim, wanted to just clarify the purpose of this PR.

There's two primary scenarios for querystring parameters with APIs that I'm aware of:

1. Querystring parameters used across all (or most) calls to the API. Examples would be parameters needed for specifying an API Version (e.g. the `api-version` parameter on calls to the [Azure REST API](https://docs.microsoft.com/rest/api/azure/#request-uri)) or some type of authentication token (e.g. something from [Amazon's S3 API](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html)). 
2. Querystring parameters that are specific to a particular API call, acting similarly to those that are part of the path of the url (the ones we can `cd` through in the tool).
3. Querystring parameters that are in common to all (or many) calls to the API, but that you wouldn't necessarily pass the same value on each call (e.g. pagination parameters)

From my initial glance at this PR, it seems like its goal is to provide an implementation for the first scenario, since the values are stored as part of the `HttpState` object:

https://github.com/dotnet/HttpRepl/blob/32f45bf63dfd576f1d33ee6b4897ce24d2053291/src/Microsoft.HttpRepl/Commands/SetQueryStringCommand.cs#L50-L59

and passed to all commands:

https://github.com/dotnet/HttpRepl/blob/32f45bf63dfd576f1d33ee6b4897ce24d2053291/src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs#L134

Is that accurate?

I'd be interested to hear what the folks from #258 (@bravecobra, @salihkarabilo, @Leon99) say about whether this would address their needs, or if they need scenarios 2 & 3 as well.

ygalbel:
Hi,
Thanks for your answer.

As you said, my code fix case number 1.

I am working against an API that get username and password as part of queryString,
so that is the reason I need to add this kind of support, and use the same queryString in all requests.

Maybe I can add the support for other scenarios if it will be used.
That don't seems to be a lot of work.

tlmii:
@ygalbel I'd say hold off on trying to implement the others. I think scenario 1 that you've already implemented is the most straightforward from a UX perspective and we probably want to give the others a bit more consideration. I'll take a pass at reviewing the PR shortly and let the others chime in when they get back.

@darrelmiller - correct me if I'm wrong, but there's nothing in the OpenAPI spec that would differentiate in the OpenAPI description whether a parameter was "global" (like scenario 1 or 3) vs specific to the particular path/method, right? It probably _should_ be split out as a `$ref` in the global scenarios, but that's not required _and_ it could be split out as a `$ref` in scenario 2 anyway. Is that correct?

src/Microsoft.HttpRepl/Commands/HelpCommand.cs
```diff
@@ -186,6 +186,7 @@ public static void CoreGetHelp(IShellState shellState, ICommandDispatcher<HttpSt
             output.AppendLine();
             output.AppendLine($"{"connect",navCommandColumn}{GetCommand<ConnectCommand>(dispatcher).GetHelpSummary(shellState, programState)}");
             output.AppendLine($"{"set header",navCommandColumn}{GetCommand<SetHeaderCommand>(dispatcher).GetHelpSummary(shellState, programState)}");
+            output.AppendLine($"{"set queryString",navCommandColumn}{GetCommand<SetQueryStringCommand>(dispatcher).GetHelpSummary(shellState, programState)}");
```
tlmii, line 189: From the output of this, it looks like this is now our longest command name, so we'll need to bump `navCommandColumn` (probably bump it by 2 since the command will be longer when it is changed to `query-string`):

```console
Setup Commands:
Use these commands to configure the tool for your API server

connect        Configures the directory structure and base address of the api server
set header     Sets or clears a header for all requests. e.g. `set header content-type application/json`
set queryStringSets or clears a query string key and value. When [value] is empty the header is cleared. The value will be UrlEncoded.
```

tlmii, line 189: `query-string`

test/Microsoft.HttpRepl.Tests/HttpStateTests.cs
```diff
@@ -169,11 +169,80 @@ public void GetEffectivePath_ProperConcatenation(string baseUriString, string pa
...
+            string specifiedPath = "dir2";
+            var queryString = new Dictionary<string, IEnumerable<string>>()
+            {
+                ["key"] = new List<string>() { "value"}
+            }; 
+            Uri baseUri = new Uri(baseUriString);
+
+            Uri result = HttpState.GetEffectivePath(baseUri, pathSections, specifiedPath, queryString);
+
+            Assert.Equal("https://github.com/dir1/dir2?key=value", result.ToString(), StringComparer.OrdinalIgnoreCase);
```
tlmii, line 191: Can you lift the string literal here to a local variable at the top (probably something like `expectedPath`) so they're near the construction of the parameters for easy comparison? Same for the other test methods below.

test/Microsoft.HttpRepl.Tests/HttpStateTests.cs
```diff
@@ -169,11 +169,80 @@ public void GetEffectivePath_ProperConcatenation(string baseUriString, string pa
...
             Assert.Equal(expectedResult, result.ToString(), StringComparer.OrdinalIgnoreCase);
         }
 
+        [Fact]
+        public void GetEffectivePath_AppendQueryString_SingleCase()
+        {
+            string baseUriString = "https://github.com/";
+            string pathSections = "dir1";
+            string specifiedPath = "dir2";
+            var queryString = new Dictionary<string, IEnumerable<string>>()
```
tlmii, line 183: Same nit about `var` usage in this file as mentioned elsewhere. Note that you can use target-typed new expression here to keep the generic `Dictionary`/`IEnumerable` syntax from being so ugly:
```C#
Dictionary<string, IEnumerable<string>> queryString = new()
{
   ...
}
```

src/Microsoft.HttpRepl/HttpState.cs
```diff
@@ -84,10 +88,16 @@ public IEnumerable<string> GetApplicableContentTypes(string method, string path)
 
         public Uri GetEffectivePath(string commandSpecifiedPath)
         {
-            return GetEffectivePath(BaseAddress, string.Join('/', PathSections.Reverse()), commandSpecifiedPath);
+            return GetEffectivePath(BaseAddress, string.Join('/', PathSections.Reverse()), commandSpecifiedPath, null);
+        }
+
+
+        public Uri GetEffectivePathWithQueryString(string commandSpecifiedPath, Dictionary<string,IEnumerable<string>> queryString)
```
tlmii, line 95: I need to give this more thought. I _think_ its more correct to include the query string by default (and force the use of the static if you want only some of the parts - as is done in `GetEffectivePathForPrompt`). But I need to dig through some of the usages first.

test/Microsoft.HttpRepl.Tests/HttpStateTests.cs
ygalbel: That didn't works for me.
I don't know why.
For now I will keep 
````
Dictionary<string, IEnumerable<string>> queryString = new Dictionary<string, IEnumerable<string>>()
`````

File: test/Microsoft.HttpRepl.Tests/HttpStateTests.cs (hunk 2 above)
Comment (ygalbel) [reply test/Microsoft.HttpRepl.Tests/HttpStateTests.cs:191]: Done

File: src/Microsoft.HttpRepl/Commands/SetQueryStringCommand.cs [hunk 5]
```diff
@@ -0,0 +1,91 @@
...
+using Microsoft.Repl.Parsing;
+
+namespace Microsoft.HttpRepl.Commands
+{
+    public class SetQueryStringCommand : ICommand<HttpState, ICoreParseResult>
+    {
+        private const string CommandName = "set";
+        private const string SubCommand = "queryString";
+
+        private readonly ITelemetry _telemetry;
```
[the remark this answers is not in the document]
ygalbel: Done

src/Microsoft.HttpRepl/Commands/SetQueryStringCommand.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        public SetQueryStringCommand(ITelemetry telemetry)
+        {
+            _telemetry = telemetry;
+        }
+
+        public bool? CanHandle(IShellState shellState, HttpState programState, ICoreParseResult parseResult)
+        {
+            return parseResult.ContainsAtLeast(minimumLength: 3, CommandName, SubCommand)
+                ? (bool?)true
```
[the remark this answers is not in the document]
tlmii: Interesting. What version of VS are you using?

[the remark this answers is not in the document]
ygalbel: Version 16.8.2.
I think is the latest.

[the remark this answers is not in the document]
tlmii: Hm. Yeah I can repro that too. 16.8.2 doesn't like it, but the 16.9 previews I have on my box do. Not sure why they'd be different here. Oh well. You can leave it how it works in 16.8.2 since that's the non-preview release.

bravecobra:
> @ygalbel Thanks for taking the time to submit the PR!
> 
> I'll want input from @bradygaster, @sayedihashimi, etc when we get past the long holiday weekend.
> 
> But in the interim, wanted to just clarify the purpose of this PR.
> 
> There's three primary scenarios for query string parameters with APIs that I'm aware of:
> 
> 1. Query string parameters used across all (or most) calls to the API. Examples would be parameters needed for specifying an API Version (e.g. the `api-version` parameter on calls to the [Azure REST API](https://docs.microsoft.com/rest/api/azure/#request-uri)) or some type of authentication token (e.g. something from [Amazon's S3 API](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html)).
> 2. Query string parameters that are specific to a particular API call, acting similarly to those that are part of the path of the url (the ones we can `cd` through in the tool).
> 3. Query string parameters that are in common to all (or many) calls to the API, but that you wouldn't necessarily pass the same value on each call (e.g. pagination parameters)
> 
> From my initial glance at this PR, it seems like its goal is to provide an implementation for the first scenario, since the values are stored as part of the `HttpState` object:
> 
> https://github.com/dotnet/HttpRepl/blob/32f45bf63dfd576f1d33ee6b4897ce24d2053291/src/Microsoft.HttpRepl/Commands/SetQueryStringCommand.cs#L50-L59
> 
> and passed to all commands:
> 
> https://github.com/dotnet/HttpRepl/blob/32f45bf63dfd576f1d33ee6b4897ce24d2053291/src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs#L134
> 
> Is that accurate?
> 
> I'd be interested to hear what the folks from #258 (@bravecobra, @salihkarabilo, @Leon99) say about whether this would address their needs, or if they need scenarios 2 & 3 as well.

Yeah, I would be requiring all three cases, however that would by scripting out the api calls, not by manually typing out each call at the command-line fortunately.

src/Microsoft.HttpRepl/HttpState.cs
```diff
@@ -99,6 +109,7 @@ public static Uri GetEffectivePath(Uri baseAddress, string pathSections, string
             {
                 throw new ArgumentNullException(nameof(baseAddress), string.Format(Resources.Strings.HttpState_Error_NoAbsoluteUriNoBaseAddress, nameof(commandSpecifiedPath), nameof(baseAddress)));
             }
+
```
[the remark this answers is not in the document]
bradygaster: thanks!

bradygaster:
Like @tlmii said, thanks for the PR, @ygalbel and for working through some suggestions. I'm eager to get support for QS variables so we unblock a lot of key scenarios, like _anyone using `[FromQuery]` out there_, so I'd argue getting support for items 1 and 2 would a must. 3 i'm less concerned about.

ygalbel:
Anything I can help, or you are taking the remaining development from here?
Do you want that all the feature will be finished before you will merge this Pull Request?

bravecobra:
Does this work with the same query string key being passed multiple times with different values to provide a list? We do have that scenario too.

ygalbel:
@bravecobra sure,
this case (`?key=value1&key=value2`) is totally supported and tested.

src/Microsoft.HttpRepl/HttpState.cs
```diff
@@ -108,6 +119,18 @@ public static Uri GetEffectivePath(Uri baseAddress, string pathSections, string
...
             AppendQueryToBuilder(builder, commandQuery);
 
+            if (queryString != null)
+            {
+                
+                foreach (KeyValuePair<string, IEnumerable<string>> tuple in queryString)
+                {
+                    foreach (var singleValue in tuple.Value)
+                    {
+                        AppendQueryToBuilder(builder, $"{tuple.Key}={WebUtility.UrlEncode(singleValue)}");
```
atifaziz, line 129:
```suggestion
                        AppendQueryToBuilder(builder, $"{WebUtility.UrlEncode(tuple.Key)}={WebUtility.UrlEncode(singleValue)}");
```

According to [“application/x-www-form-urlencoded “](https://www.w3.org/TR/html401/interact/forms.html#didx-applicationx-www-form-urlencoded):

> 1. Control names and values are escaped.

It's also best to err on the safe side. I'd also recommend adding a test for this.

Also, any considerations for the following?

> Additionally, the `allowReserved` keyword specifies whether the reserved characters `:/?#[]@!$&'()*+,;=` in parameter values are allowed to be sent as they are, or should be percent-encoded. By default, `allowReserved` is `false`, and reserved characters are percent-encoded. For example, `/` is encoded as `%2F` (or `%2f`), so that the parameter value `quotes/h2g2.txt` will be sent as `quotes%2Fh2g2.txt`

src/Microsoft.HttpRepl/Resources/Strings.resx
```diff
@@ -434,4 +437,4 @@ Read more about .NET CLI Tools telemetry: https://aka.ms/dotnet-cli-telemetry
   <data name="ApiConnection_Logging_SuccessfulWithWarnings" xml:space="preserve">
     <value>Successful (with warnings)</value>
   </data>
-</root>
\ No newline at end of file
+</root>
```
atifaziz, line 440: nit: unrelated change

src/Microsoft.HttpRepl/HttpState.cs
```diff
@@ -108,6 +119,18 @@ public static Uri GetEffectivePath(Uri baseAddress, string pathSections, string
             AppendQueryToBuilder(builder, baseAndPathQuery);
             AppendQueryToBuilder(builder, commandQuery);
 
+            if (queryString != null)
+            {
+
```
atifaziz, line 124: nit: extra line:

```suggestion
```

src/Microsoft.HttpRepl/HttpState.cs
```diff
@@ -84,10 +88,16 @@ public IEnumerable<string> GetApplicableContentTypes(string method, string path)
 
         public Uri GetEffectivePath(string commandSpecifiedPath)
         {
-            return GetEffectivePath(BaseAddress, string.Join('/', PathSections.Reverse()), commandSpecifiedPath);
+            return GetEffectivePath(BaseAddress, string.Join('/', PathSections.Reverse()), commandSpecifiedPath, null);
+        }
+
```
atifaziz, line 93: nit: extra blank line:

```suggestion
```

atifaziz (changes requested):

src/Microsoft.HttpRepl/Commands/HelpCommand.cs
tlmii: I definitely want to pull in @bradygaster on this one. I agree with the logic of what you said - including the query-param part... but I do wonder a bit about user expectations. Does it matter if (and this is just a guess on my part) our users would be more familiar with the term "query string" being used colloquially to cover a single query parameter? It seems like `query`/`query-string` would be aliases for each other (if we add those commands) and would be things the user might try on their own, but `query-param` might not be as obvious. I'd be happy to be wrong there, though, as I certainly prefer the precise language suggested here.

src/Microsoft.HttpRepl/HttpState.cs
```diff
@@ -34,6 +35,8 @@ public class HttpState
 
         public Dictionary<string, IEnumerable<string>> Headers { get; }
 
+        public Dictionary<string, IEnumerable<string>> QueryString{ get; }
```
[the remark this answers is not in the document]
tlmii: > Now if the scope of HTTP REPL is just web APIs per Open API

That's certainly a/the focus, but I wouldn't want to constrain us to that when there's a relatively easy way to support more scenarios. 

That said, I see a bit of a problem here in that the list being created here (global query parameters, sent to all subsequent HTTP commands) will be combined with per-command query parameters specified at the time of the specific HTTP command.

It would seem that the best we could do would be to guarantee 
- Order within global query parameters was maintained
- Order within command-specific query parameters was maintained
- Global query parameters are always listed... before? after? command-specific query parameters

I suspect that the situation where order is important AND the user would prefer to have a mixture of global and command-specific parameters would be exceedingly rare, but I wanted to call it out here in case anyone has any thoughts.

src/Microsoft.HttpRepl/HttpState.cs
tlmii: re: `allowReserved`:  I'd be curious to know how often allowReserved is true on query parameters (a quick [search of APIs.guru's OpenAPI Directory](https://github.com/APIs-guru/openapi-directory/search?q=allowReserved) turned up only 1 result, though I doubt that's fully representative. My instinct says we should table that and add an issue to track supporting it later, in large part because that's a per-endpoint decision and the parameter definitions aren't available at the time this code is being called, thus making it a larger change to support it. Seems like a good candidate for its own issue/PR.

src/Microsoft.HttpRepl/Commands/HelpCommand.cs
atifaziz: Folks, during a recent GitHub service hiccup/incident, my original comment got deleted. The support staff at GitHub were kind enough to to recover the comment body, which I posting back here for context and belong right before @tlmii reply above:

> Why not make it simpler and call it `query`? Does the string part add or help anything except more typing? It would align nicely with [Section 3, “Syntax Components”](https://tools.ietf.org/html/rfc3986#section-3) (as well as [sub-section 3.4](https://tools.ietf.org/html/rfc3986#section-3.4)) of [RFC 3986](https://tools.ietf.org/html/rfc3986#section-3) where it's called just _query_:
> 
>     foo://example.com:8042/over/there?name=ferret#nose
>     \_/ \______________/\_________/ \_________/ \__/
>     | | | | |
>     scheme authority path query fragment
>     | _____________________|__
>     / \ / \
>     urn:example:animal:ferret:nose
> 
> If anything, `query-string` should be an alias like `dir` is for `ls`.
> 
> That said, there is another problem. The query component is opaque. The fact that it contains name-value pairs came from [URL-encoding of HTML forms](https://www.w3.org/TR/html401/interact/forms.html#h-17.13.3.4) and got adopted more widely outside the context of HTML form submission. The name-value pairs have become more conventionally known as _parameters_, e.g. [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/get#Examples) and [query parameters in Open API](https://swagger.io/docs/specification/describing-parameters/#query-parameters), so I would argue that this should technically be called `query-param`. This way, `query`/`query-string` could be reserved to parsing and setting parameters given an entire “query string”. In other words, doing `set query foo=1&bar=2` would be equivalent to doing `set query-param foo 1` and `set query-param bar 2`.

tlmii:
@ygalbel Now that #460 is in, which (by default) clears out headers and the path when running `connect`, we should update that in this PR to clear the global query parameters and add a `--persist-query-params` similar to the `--persist-headers` and `--persist-path` that were added in that PR.

src/Microsoft.HttpRepl/Commands/HelpCommand.cs
tlmii: @ygalbel In the interest of moving this along (and since we won't push out a non-preview 5.1 release with this included for a while anyway) I'm going to go ahead and say we should go with what @atifaziz said: Make the command be `set query-param`. For now, lets leave it at that, without any additional `set query` or `set query-string` or aliases for any of it. We can see what the feedback is for the previews and get some more input from @bradygaster and others.

src/Microsoft.HttpRepl/HttpState.cs
tlmii: @ygalbel I'll add a tracking issue for the `allowReserved` portion of this discussion and mention it here when I do. Then I think you just need the change @atifaziz mentioned suggested to encode the name as well as the value. A test would be good too :)

ygalbel:
Don't remember where it stands.
Tag me if you need help from my side.

tlmii:
We were actually just discussing this internally a couple of days ago. I'm going to try to pick it back up and get it into the product here in the next few days. It looks like I've let it go stale long enough I might need to do a fresh PR to get around some of the conflicts and changes that have happened in the interim. Regardless, should be moving forward soon.

ygalbel:
Sure. Go ahead it's all yours.

tlmii:
@ygalbel @bravecobra - OK, I've created a new PR: #611 that updates everything I could see that was unresolved above, renames a bunch of things, adds one more test, fixes an integration test that broke somewhere in the various .NET/kestrel upgrades but otherwise is fully the work above. Let me know over there what you think.

tlmii:
Closing this since #611 went in. Thanks again @ygalbel for your help here!

PR #460, opened 2020-12-14
Clear state on connect

### LINKED ISSUE #449: connect command should clear session settings by default
Currently, if you `connect` to one host, and drill down into the API, then `connect` to a different one, the base address will change but the relative path beyond that won't. Additionally, the stored headers and, soon, the stored query string parameters, won't change. I think this would be unexpected to most users. That said, there are cases where you might want to maintain it - for instance switching between dev/prod hosts. So we could add an option to preserve it.

tlmii:
Closes #449

### DIFF
--- a/src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
+++ b/src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
@@ -205,7 +205,7 @@ namespace Microsoft.HttpRepl.Commands
                     SwaggerUri = programState.SwaggerEndpoint,
                     AllowBaseOverrideBySwagger = false
                 };
-                await apiConnection.SetupHttpState(programState, performAutoDetect: false, cancellationToken).ConfigureAwait(false);
+                await apiConnection.SetupHttpState(programState, performAutoDetect: false, persistHeaders: true, persistPath: true, cancellationToken).ConfigureAwait(false);
             }
         }
 
--- a/src/Microsoft.HttpRepl/Commands/ListCommand.cs
+++ b/src/Microsoft.HttpRepl/Commands/ListCommand.cs
@@ -50,7 +50,7 @@ namespace Microsoft.HttpRepl.Commands
                         SwaggerUri = programState.SwaggerEndpoint,
                         AllowBaseOverrideBySwagger = false
                     };
-                    await apiConnection.SetupHttpState(programState, performAutoDetect: false, cancellationToken).ConfigureAwait(false);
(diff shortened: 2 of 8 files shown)


src/Microsoft.HttpRepl/Commands/ListCommand.cs
```diff
@@ -50,7 +50,7 @@ protected override async Task ExecuteAsync(IShellState shellState, HttpState pro
                         SwaggerUri = programState.SwaggerEndpoint,
                         AllowBaseOverrideBySwagger = false
                     };
-                    await apiConnection.SetupHttpState(programState, performAutoDetect: false, cancellationToken).ConfigureAwait(false);
+                    await apiConnection.SetupHttpState(programState, performAutoDetect: false, persistHeaders: true, persistPath: true, cancellationToken).ConfigureAwait(false);
```
bzhu94, line 53: Not too familiar with the `auto` flag for the list command, but would it ever make sense to include the options to persist headers / path here?

src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
```diff
@@ -205,7 +205,7 @@ private async Task CreateDirectoryStructureForSwaggerEndpointAsync(IShellState s
                     SwaggerUri = programState.SwaggerEndpoint,
                     AllowBaseOverrideBySwagger = false
                 };
-                await apiConnection.SetupHttpState(programState, performAutoDetect: false, cancellationToken).ConfigureAwait(false);
+                await apiConnection.SetupHttpState(programState, performAutoDetect: false, persistHeaders: true, persistPath: true, cancellationToken).ConfigureAwait(false);
```
bzhu94, line 208: Same question as for `ListCommand` - does it make sense to check the `commandInput` param to see if the user wants to persist headers / path?

src/Microsoft.HttpRepl/Commands/ListCommand.cs
tlmii: The auto flag basically just checks the swagger/openapi description again before executing the command. This is to make sure its working on the latest info. In case, for instance, you're doing this as part of your inner dev loop, you make a change to an API and you want to test it again, you don't have to connect again (to re-read the swagger/open description) - it does that for you. So for that case, I believe it makes sense to always persist the state - you don't want to lose the state during that process.

src/Microsoft.HttpRepl/Commands/BaseHttpCommand.cs
tlmii: Same answer as above.

bzhu94 (approved):
Looks good!

Issue #119, opened 2019-07-26
Simplify processing of Unhandled keystrokes


tlmii:
Currently, the `InputManager` uses a bunch of mappings in the static `KeyHandlers` class to make sure that unrecognized keystrokes (for instance, the media keys, the browser control keys, etc) don't cause anything to print in the REPL. For instance:

https://github.com/aspnet/HttpRepl/blob/d2e744a42b7a1b66fbd9a7eeef6d3e6170931815/src/Microsoft.Repl/Input/KeyHandlers.cs#L71

This became apparent when trying to move the `Microsoft.Repl` project to `netstandard1.3`, where most of those `ConsoleKey` values don't exist (they're in netstandard2.0 and in netcoreapp2.X and above). 

There are other keys that aren't in the current `ConsoleKey` and would cause this, they're just rare. But the `netstandard1.3` change makes that set of keys much larger and more likely to be hit (even accidentally). 

We need a better way to handle the full set of otherwise-unhandled unprintable characters. As I mentioned in https://github.com/aspnet/HttpRepl/pull/117#discussion_r307539557, one option would be to have our own `char.IsPrintable( ... )` implementation, based on something like `IsLetterOrDigit || IsPunctuation || IsSymbol || Environment.NewLine || ' '` but we would need to take care to get that right. Once we had that, if the `ConsoleKey` wasn't handled by `KeyHandlers`, and the character for that key didn't meet our definition of printable, then we could safely just ignore it.

I'm putting this issue in here to track this work since I think it is separate from the conversion of the library to netstandard1.3, but still very important.

sayedihashimi:
**Thank you for your interest in HttpRepl.**

This project has been deprecated, and the repository is now archived. As part of this transition, we are closing all open issues.
Please read the https://github.com/dotnet/HttpRepl/issues/701 for alternative recommendations.

PR #475, opened 2021-01-12
Drop null character-causing keystrokes

### LINKED ISSUE #472: Several characters are count twice
I'm running `5.0.2-preview.21057.1`, but I'm pretty sure this was also the case for `5.0.1`

There are several characters that count twice in terms of cursor movement.

Steps to reproduce:
- Type `C:\`
- Press `<Home>`
- Press `<End>`
- Press `<Backspace>` five times

Expected behavior: The cursor should be right before the `C` after `<Home>` and `<Backspace>` should remove three characters
Actual behavior: The cursor is within the prompt before the `>` and you can remove five characters including two from the prompt.

Steps to reproduce:
- Type `:\:\:\:\:\:\:\:\` (That is as many characters as there are in the `(Disconnected)> ` prompt)
- Press `<Backspace>` 32 times (or just hold it)

Expected behavior: The 16 characters should be removed
Actual behavior: The 16 characters as well as the 16 that form the prompt are removed and the prompt is invisible

httprepl-characters-twice

tlmii:
Resolves #472

This handles a specific case of #119 where the unprintable character coming through is the null character. This seems common enough (perhaps the most common of the scenarios highlighted by #119) to handle on its own.

### DIFF
--- a/src/Microsoft.Repl/Input/InputManager.cs
+++ b/src/Microsoft.Repl/Input/InputManager.cs
@@ -337,6 +337,15 @@ namespace Microsoft.Repl.Input
                     }
                     else
                     {
+                        // If we got here, we've processed all handlers we know for
+                        // key combinations. So anything else should have a valid
+                        // character or be the null character. If its the latter,
+                        // we just want to ignore it.
+                        if (keyPress.KeyChar == '\0')
+                        {
+                            continue;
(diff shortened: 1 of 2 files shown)


bzhu94 (approved):
Looks good!

PR #487, opened 2021-02-11
Switch to Directory.Packages.props and update one dependency

tlmii:
Needed to update a dependency in the test project for a security vulnerability, so I went ahead and switched to using Directory.Packages.props as well.


jimmylewis (approved):
LGTM

PR #533, opened 2021-11-16
Switch to servicing pools on release/6.0

jonfortescue:


tlmii (dismissed):

tlmii:
@jonfortescue I hit approve on this, but then noticed it was targeted at main, not release/6.0. I assume that wasn't intentional?

Also, just to confirm, making the changes to the arcade-supplied .yml files is ok to do (I generally don't touch anything in `/eng/common`.

jonfortescue:
Ah thank you for the good catch!

jonfortescue:
@tlmii Could you take another look at this?

tlmii (approved):

PR #557, opened 2022-06-04
Remove .NET 5.0 as it is out of support

tlmii:
.NET 5.0 reached [end of support](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core) last month, so we'll remove it as a target framework.


bzhu94 (approved):
LGTM

PR #593, opened 2022-12-01
Update System.Drawing.Common to avoid vulnerability

tlmii:
We got a CG alert for `System.Drawing.Common` 4.7.0. Eventually we can update the AppInsights dependency to 2.22 and it'll bring in something without the vulnerability, but for now we need to pin `System.Drawing.Common` to 4.7.2.

### DIFF
--- a/src/Microsoft.HttpRepl.Telemetry/Microsoft.HttpRepl.Telemetry.csproj
+++ b/src/Microsoft.HttpRepl.Telemetry/Microsoft.HttpRepl.Telemetry.csproj
@@ -7,4 +7,13 @@
     <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" />
     <PackageReference Include="Microsoft.Win32.Registry" />
   </ItemGroup>
+
+  <ItemGroup>
+    <!--
+      This PackageReference is only here to deal with a Component Governance alert.
+      Once Microsoft.ApplicationInsights.WorkerService 2.22 is released, we can
+      update our reference to that and remove this.
+    -->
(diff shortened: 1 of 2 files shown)


src/Microsoft.HttpRepl.Telemetry/Microsoft.HttpRepl.Telemetry.csproj
```diff
@@ -7,4 +7,13 @@
...
     <PackageReference Include="Microsoft.Win32.Registry" />
   </ItemGroup>
+
+  <ItemGroup>
+    <!--
+      This PackageReference is only here to deal with a Component Governance alert.
+      Once Microsoft.ApplicationInsights.WorkerService 2.22 is released, we can
+      update our reference to that and remove this.
+    -->
+    <PackageReference Include="System.Drawing.Common" />
```
jimmylewis, line 17: This should be unnecessary.  With CPM, versions pinned in the Directory.Packages.props file are applied to transitive dependencies as well.

tlmii: I didn't have transitive pinning enabled yet (its disabled by default). Added it, removed this PackageReference and a couple others that were just for transitive dependencies.

jimmylewis: Oh right, I forgot it's not the default (it was during the feature previews).  LGTM now ;)

jimmylewis (approved):

PR #594, opened 2022-12-01
Add net7.0, remove netcoreapp3.1 and netstandard2.0

### LINKED ISSUE #592: Cannot run Microsoft.dotnet-httprepl 6.0.0 with dotnet 7.0 (at Ubuntu 22.04)
Microsoft.dotnet-httprepl 6.0.0 does work with dotnet 6.0 as so far, but it seems not compatible with dotnet 7.0.
Here is detail message:

You must install or update .NET to run this application.

App: /home/ubuntu/.dotnet/tools/httprepl
Architecture: x64
Framework: 'Microsoft.NETCore.App', version '6.0.0' (x64)
.NET location: /usr/share/dotnet

The following frameworks were found:
  7.0.0 at [/usr/share/dotnet/shared/Microsoft.NETCore.App]

Learn about framework resolution:
https://aka.ms/dotnet/app-launch-failed

How should I fix this issue, I really need httprepl for our production now.
Thanks in advance !

tlmii:
.NET 7 has been released and .NET Core 3.1 goes out of support in less than 2 weeks. So here is a PR to take care of both of those things.

Since we're dropping .NET Core 3.1 support, we'll also move the usage of .NET Standard 2.0 to .NET 6. That will eventually enable us to take advantage of some things that .NET Standard prevented us from doing. For now, the only additional changes are to address analyzers after the update. That includes bringing DockerContainerDetectorForTelemetry up to date with what's in the SDK itself. 

In keeping with past .NET releases, we're also bumping the version of the tool to 7.0

Resolves #592


src/Microsoft.HttpRepl.Telemetry/DockerContainerDetectorForTelemetry.cs
```diff
@@ -1,51 +1,67 @@
...
-                        : IsDockerContainerResult.False;
-                case Platform.Unknown:
-                    return IsDockerContainerResult.Unknown;
-                case Platform.Darwin:
-                default:
-                    return IsDockerContainerResult.False;
+                }
+                catch (SecurityException)
+                {
+                    return HttpRepl.Telemetry.IsDockerContainer.Unknown;
```
jimmylewis, line 30: nit: inconsistent style here (catch and return Unknown) vs the IsLinux() branch (catch and fall out to the default) for basically the same structure.

jimmylewis (approved):

PR #610, opened 2023-03-01
Update version to 7.1

tlmii:


jimmylewis:
Just curious, is there a reason for this bump?  Or is it just maintaining alignment with something else?

tlmii:
> Just curious, is there a reason for this bump? Or is it just maintaining alignment with something else?

Going to finally take the `set query-param` PR, and since that's a new command, it seemed necessary to bump it.

PR #611, opened 2023-03-01
Add set query-param command

tlmii:
This PR is based on #448 by @ygalbel but because I let that PR go so stale, it was difficult to rebase and rework it. I've included the suggestions on the PR from @atifaziz in terms of the naming and url encoding, I added another test and fixed some things that broke locally in the integration tests.

### DIFF
--- a/src/Microsoft.HttpRepl/Commands/SetQueryParamCommand.cs
+++ b/src/Microsoft.HttpRepl/Commands/SetQueryParamCommand.cs
@@ -1,0 +1,91 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the License.txt file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.HttpRepl.Resources;
+using Microsoft.HttpRepl.Telemetry;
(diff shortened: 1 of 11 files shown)


src/Microsoft.HttpRepl/Commands/SetQueryParamCommand.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        public string Name => "setQueryParam";
+        public static string Description => Strings.SetQueryParamCommand_HelpDetails;
+
+        public SetQueryParamCommand(ITelemetry telemetry)
+        {
+            _telemetry = telemetry;
+        }
+
+        public bool? CanHandle(IShellState shellState, HttpState programState, ICoreParseResult parseResult)
```
bradygaster, line 36: couldn't this just be a => ?

src/Microsoft.HttpRepl/Commands/SetQueryParamCommand.cs
```diff
@@ -0,0 +1,91 @@
...
+                helpText.AppendLine("set query-param {name} [value]");
+                helpText.AppendLine();
+                helpText.AppendLine(Strings.SetQueryParamCommand_HelpDetails);
+                return helpText.ToString();
+            }
+
+            return null;
+        }
+
+        public string GetHelpSummary(IShellState shellState, HttpState programState)
```
bradygaster, line 81: And also these two?

bradygaster (approved):

PR #614, opened 2023-04-20
add query-param and clear query-param command

bleaphar:
This pr is a rework of [611](https://github.com/dotnet/HttpRepl/pull/611) to split up the logic for adding to query-param and clearing query-param as well as messaging to alert the user if a value is not mapped to a key as add query-param now requires keys and values to be set up in pairs.


.config/dotnet-tools.json
```diff
@@ -0,0 +1,12 @@
+{
+  "version": 1,
+  "isRoot": true,
+  "tools": {
+    "microsoft.dotnet-httprepl": {
+      "version": "7.0.0",
+      "commands": [
+        "httprepl"
```
tlmii, line 8: You probably don't want this file checked in (or at least not with httprepl in it)

src/Microsoft.HttpRepl/HttpState.cs
```diff
@@ -117,7 +135,7 @@ public Uri GetEffectivePathForPrompt()
                 return null;
             }
 
-            return GetEffectivePath(BaseAddress, string.Join('/', PathSections.Reverse()), "");
+            return GetEffectivePath(BaseAddress, string.Join('/', PathSections.Reverse()), "", queryParam: null);
```
tlmii, line 138: I wonder about whether we should display the query params for the prompt or not (currently it does not). But if it is not displayed on the prompt, then you can't really see what the actual URL is that will be executed if you do a GET or POST. 

Perhaps we could defer that and have a separate PR that made it an option for whether or not to display query params.

src/Microsoft.HttpRepl/HttpState.cs
```diff
@@ -83,10 +85,15 @@ public IEnumerable<string> GetApplicableContentTypes(string method, string path)
 
         public Uri GetEffectivePath(string commandSpecifiedPath)
         {
-            return GetEffectivePath(BaseAddress, string.Join('/', PathSections.Reverse()), commandSpecifiedPath);
+            return GetEffectivePath(BaseAddress, string.Join('/', PathSections.Reverse()), commandSpecifiedPath, queryParam: null);
+        }
+
+        public Uri GetEffectivePathWithQueryParam(string commandSpecifiedPath, Dictionary<string, IEnumerable<string>> queryParam)
```
tlmii, line 91: A few things seem wrong here:

1. We have GetEffectivePathWithQueryParam that takes the queryParam parameter when the only call into it uses the query params that are stored in this object.
2. We have three public GetEffectivePath* methods, even though only one is used outside of this class (or test cases).
3. The one GetEffectivePath* method that is used outside of this class is the lengthier-named one.

So I have a suggestion. What if we instead had (psuedo-code):

```C#
private static Uri GetEffectivePath(Uri baseAddress, string pathSections, string commandSpecifiedPath, Dictionary<string, IEnumerable<string>> queryParam)
{
    // The same as the current static method
}

private Uri GetEffectivePathWithoutQueryParam(string commandSpecified)
    => GetEffectivePath(BaseAddress, string.Join('/', PathSections.Reverse()), commandSpecifiedPath, queryParam: null);

public Uri GetEffectivePath(string commandSpecified)
    => GetEffectivePath(BaseAddress, string.Join('/', PathSections.Reverse()), commandSpecifiedPath, QueryParam);
```

(private may need to be internal for tests)

And then change the current calls to GetEffectivePath and GetEffectivePathWithQueryParam to the appropriate new one to match the current behavior (or adjusted behavior per other comments).

Thoughts?

tlmii (changes requested):

src/Microsoft.HttpRepl/HttpState.cs
bleaphar: I like the thought of separating these concerns and creating an issue with it and I'll get to implementing a sort of optional echo feature.

src/Microsoft.HttpRepl/HttpState.cs
bleaphar: Sounds good to me, the latest commit addresses these changes!

tlmii (approved):

PR #626, opened 2023-11-04
Delete useless null check for CommandTestsBase

kurnakovv:
Hi! Thank you for this tool! ;)

I read code and found duplicate null check. As you can see, [this](https://github.com/dotnet/HttpRepl/blob/main/test/Microsoft.HttpRepl.Tests/Commands/CommandTestsBase.cs#L125) null check is useless, because [here](https://github.com/dotnet/HttpRepl/blob/main/test/Microsoft.HttpRepl.Tests/Commands/CommandTestsBase.cs#L121) we already check path by null

Hope my changes make your code better ;)


tlmii (approved):

tlmii:
Thanks for the PR!

PR #682, opened 2024-10-24
Remove net6.0 and net7.0 versions from shipping packages

tlmii:
Dropping net6/net7 builds from the nuget package as they're out of support (or will be by the time we ship this in a stable build).

net7.0 is out of support as of May 14, 2024
net6.0 will be going out of support on November 12, 2024

Bumping the version to 8.1 rather than just 8.0.1 since this is a breaking change for anyone trying to run it on those runtimes.

### DIFF
--- a/src/Microsoft.Repl/Microsoft.Repl.csproj
+++ b/src/Microsoft.Repl/Microsoft.Repl.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net8.0</TargetFrameworks>
+    <TargetFramework>net8.0</TargetFramework>
     <Description>A framework for creating REPLs in .NET Standard.</Description>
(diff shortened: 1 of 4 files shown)


src/Microsoft.Repl/Microsoft.Repl.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>net8.0</TargetFrameworks>
```
deepchoudhery, line 4: nit : `TargetFramework` instead of `TargetFrameworks`

tlmii: Whoops. Good catch. Changed.

deepchoudhery (approved):
LGTM!, will check for dotnet/scaffolding as well

